Issue #6, opened 2019-06-15
floating point numbers generate ERROR nodes and trunk values


cancerberoSgx:
The following code: 
```scala
val v = 3.14
```
generates the following AST: 
```json
{"type":"compilation_unit","text":"val v = 3.14","children":[{"type":"val_definition","text":"val v = 3.14","children":[{"type":"val","text":"val","children":[]},{"type":"identifier","text":"v","children":[]},{"type":"=","text":"=","children":[]},{"type":"ERROR","text":"3.","children":[{"type":"number","text":"3","children":[]},{"type":".","text":".","children":[]}]},{"type":"number","text":"14","children":[]}]}]}
```

(Executing tree.rootNode.toString() gives `(compilation_unit (val_definition (identifier) (ERROR (number)) (number)))`

Seems the problem is with float numbers. The number is divided in three nodes, so it's not possible to access its value.  I think it should behave just like with integers, so I would expect: 

```
{"type":"compilation_unit","text":"val v = 3.14","children":[{"type":"val_definition","text":"val v = 3.14","children":[{"type":"val","text":"val","children":[]},{"type":"identifier","text":"v","children":[]},{"type":"=","text":"=","children":[]},{"type":"number","text":"3.14","children":[]}]}]}
```

PR #7, opened 2019-06-15
Add basic floating point literal parsing

cancerberoSgx:
*  https://github.com/tree-sitter/tree-sitter-scala/issues/6
* updates dependencies (so wasm can be generated)
* probably there is a better way of doing this, I tried 1) copy/ paste from Java and 2) trying to import as external from node-types.json. But none didn't seem to work 

Unfortunately I don't have much time to invest on this. Nevertheless I think this patch at least alleviates the problem. If you merge it, could you please also npm publish it ? 

Thanks


maxbrunsfeld:
Thanks! This will at least handle most floating point values.

PR #11, opened 2020-06-29
Update readme and run build

ckipp01:
I'm been curiously watching this project for a while, and I'm attempting to take a stab at a couple things. This just runs build and updates some dead/old links in the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,8 +7,7 @@ Scala grammar for [tree-sitter](https://github.com/tree-sitter/tree-sitter).
 
 References
 
-* [The Scala Language Specification](http://www.scala-lang.org/docu/files/ScalaReference.pdf)
-* [Scala Syntax Summary](https://www.scala-lang.org/files/archive/spec/2.11/13-syntax-summary.html)
+* [Scala Syntax Summary](https://scala-lang.org/files/archive/spec/2.13/13-syntax-summary.html)
 
 Development
 -----------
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -7,8 +7,7 @@ Scala grammar for [tree-sitter](https://github.com/tree-sitter/tree-sitter).
 
 References
 
-* [The Scala Language Specification](http://www.scala-lang.org/docu/files/ScalaReference.pdf)
```
ckipp01, line 10: I'm not sure what happened to this, but it's not there anymore and I couldn't easily find what replaced it. For now I figured it's best to just remove it.

README.md
```diff
@@ -7,8 +7,7 @@ Scala grammar for [tree-sitter](https://github.com/tree-sitter/tree-sitter).
...
-* [Scala Syntax Summary](https://www.scala-lang.org/files/archive/spec/2.11/13-syntax-summary.html)
```
ckipp01, line 11: I'm going to assume that when work was starting on this that it was still in 2.11 time. There has been quite a few updates since then, so probably best to update this, especially seeing that Dotty is around the corner.

src/parser.c
```diff
@@ -1138,6 +1138,24 @@ static TSSymbol ts_alias_sequences[4][MAX_ALIAS_SEQUENCE_LENGTH] = {
   },
 };
 
+static inline bool sym_operator_identifier_character_set_1(int32_t lookahead) {
```
ckipp01, line 1141: These were just changes that happened when I ran `npm run build` which I'm assuming is due to the newer version of tree-sitter that was pulled down.

ckipp01:
Superceded by #13

PR #13, opened 2020-07-12
Various improvements

ubolonton:
- Better type support
- Annotations
- Multiple parameter lists
- Better pattern precedence
- Supertypes and CLI metadata
- Some other small improvements


maxbrunsfeld:
Thanks! This is great.

PR #16, opened 2021-03-07
Allow for catch and finally in try expression to be on a new line

ckipp01:
I noticed that `try_expression`s weren't quite looking right when they were on a new line. However, I'm having a hard time understanding exactly how to actually take the new line into account.

To give an example.

```scala
def main() = {
  try {
   doThing()
  } catch {
    case _: Throwable => 3
  }
}
```

This parses correctly and the `finally_clause` is recognized as part of the `try_expression`.

```
(compilation_unit [0, 0] - [7, 0]
  (function_definition [0, 0] - [6, 1]
    name: (identifier [0, 4] - [0, 8])
    parameters: (parameters [0, 8] - [0, 10])
    body: (block [0, 13] - [6, 1]
      (try_expression [1, 2] - [5, 3]
        body: (block [1, 6] - [3, 3]
          (call_expression [2, 3] - [2, 12]
            function: (identifier [2, 3] - [2, 10])
            arguments: (arguments [2, 10] - [2, 12])))
        (catch_clause [3, 4] - [5, 3]
          (case_block [3, 10] - [5, 3]
            (case_clause [4, 4] - [5, 2]
              pattern: (typed_pattern [4, 9] - [4, 21]
                pattern: (wildcard [4, 9] - [4, 10])
                type: (type_identifier [4, 12] - [4, 21]))
              body: (number [4, 25] - [4, 26]))))))))
```

 However, if you simply bump the `catch` down a line to create:

```scala
def main() = {
  try {
   doThing()
  }
  catch {
    case _: Throwable => 3
  }
}
```

I then totally misses the `catch_clause`.

```
(compilation_unit [0, 0] - [8, 0]
  (function_definition [0, 0] - [7, 1]
    name: (identifier [0, 4] - [0, 8])
    parameters: (parameters [0, 8] - [0, 10])
    body: (block [0, 13] - [7, 1]
      (try_expression [1, 2] - [3, 3]
        body: (block [1, 6] - [3, 3]
          (call_expression [2, 3] - [2, 12]
            function: (identifier [2, 3] - [2, 10])
            arguments: (arguments [2, 10] - [2, 12]))))
      (ERROR [4, 2] - [4, 7]
        (identifier [4, 2] - [4, 7]))
      (case_block [4, 8] - [6, 3]
        (case_clause [5, 4] - [6, 2]
          pattern: (typed_pattern [5, 9] - [5, 21]
            pattern: (wildcard [5, 9] - [5, 10])
            type: (type_identifier [5, 12] - [5, 21]))
          body: (number [5, 25] - [5, 26]))))))
```

This happen when using `finally` as well on a newline.

So this seems to make sense to me: 
https://github.com/tree-sitter/tree-sitter-scala/blob/262797b1dfe0303818c2418c0a88f6be65f37245/grammar.js#L540-L545

Although I'm not really sure how that works with newlines. Do you maybe have any pointers or insights into what I'm missing here? I'd love to get a better understanding of what's going on here and start tackling some of these missing features for the Scala grammar. I'm increasingly wanting a complete grammar as all the great stuff around Neovim and tree-sitter is happening, and I feel like Scala is missing out!


maxbrunsfeld:
I think that what's needed is a change to the external scanner's logic for inserting [`automatic_semicolon`](https://github.com/tree-sitter/tree-sitter-scala/blob/262797b1dfe0303818c2418c0a88f6be65f37245/grammar.js#L673) tokens, implemented [here](https://github.com/tree-sitter/tree-sitter-scala/blob/262797b1dfe0303818c2418c0a88f6be65f37245/src/scanner.c#L73-L93).

We need to do the same thing for `catch` and `finally` that is already done for `else`: prevent automatic semicolons for lines that begin with `else`/`finally`/`catch`.

maxbrunsfeld:
[This](https://stackoverflow.com/questions/43778992/scala-semicolon-inference-example) describes Scala's rules for automatic semicolon insertion.

> In short, a line ending is treated as a semicolon unless one of the following conditions is true:
> 1. The line in question ends in a word that would not be legal as the end of a statement, such as a period or an infix operator.
> 2. The next line begins with a word that cannot start a statement.
> 3. The line ends while inside parentheses(...) or brackets[...], because these cannot contain multiple statements anyway.

ckipp01:
Cool, thanks a lot for the pointers @maxbrunsfeld. I think this should be ready to go!

src/scanner.c
```diff
@@ -89,6 +91,40 @@ bool tree_sitter_scala_external_scanner_scan(void *payload, TSLexer *lexer,
...
+      advance(lexer);
+      if (lexer->lookahead != 't') return true;
+      advance(lexer);
+      if (lexer->lookahead != 'c') return true;
+      advance(lexer);
+      if (lexer->lookahead != 'h') return true;
+      advance(lexer);
+      if (iswalpha(lexer->lookahead)) return true;
+      return false;
+    }
```
maxbrunsfeld, line 107: Just checking - Is there ever a case where both `catch` and `finally` are valid at the same time?

ckipp01: Do you mean valid interchangably?

Like:

```scala
try doTHing()
<here> <- both catch or finally are valid here
```

Is that way you mean?

maxbrunsfeld: Yeah. So, after checking the grammar, I see that after the "body", both keywords `catch` and `finally` would be valid (because the catch clause is optional):

```
    try_expression: $ => prec.right(seq(
      'try',
      field('body', $._expression),
      optional($.catch_clause),
      optional($.finally_clause)
    )),
```

So I think that, if the keyword `catch` is valid, you still need to also check for the presence of the word `finally`.

maxbrunsfeld: Right now, if `catch` is valid, and the lookahead is not the letter `c`, the scanner will return `true` (creating an automatic semicolon). This isn't quite right, because we still want to make sure that the next keyword isn't "finally".

ckipp01: Ahhh ok, yea that totally makes sense. Alright, I accounted for that now.

ckipp01:
I'm not sure if this isn't trigger the build. It may be since this was already opened when the actions pr was merged that it doesn't register. Do you by chance want me to close this and reopen to try and trigger it?

.github/workflows/ci.yml
```diff
@@ -2,7 +2,8 @@ name: Build/test
 on:
   push:
     branches:
-      - "**"
+      - master
+    pull_request:
```
ckipp01, line 6: I took this from how we do it in Metals. https://github.com/scalameta/metals/blob/6001b60afb2774c342a73fea4856d42ff5fa5d1e/.github/workflows/ci.yml#L4-L6

maxbrunsfeld:
Thanks!

PR #20, opened 2021-03-12
Enhance/fix string interpolation model

### LINKED ISSUE #14: Change the way `interpolation` is modeled.
I wonder if the `string_transform_expression` may be better named as `processed_string_expressions`. It also seems that the parsing now always just thinks that a `$` in a literal string is interpolation, which in incorrect. According to the [spec](https://scala-lang.org/files/archive/spec/2.13/01-lexical-syntax.html#string-literals):

> Interpolated string consist of an identifier starting with a letter immediately followed by a string literal.

We also don't actually have any tests in corpus covering `string_transform_expression`. With the way current things are, the following would parse like so:

```
==========================
Interpolated string
==========================

val stringLiteral = "$a"
---

(compilation_unit
  (val_definition (identifier)
    (string (interpolation (identifier)))))
```

When in reality, I think this should just be:

```
==========================
Interpolated string
==========================

val stringLiteral = "$a"
---

(compilation_unit
  (val_definition (identifier) (string)))
```

Currently with the `string_transorm_expression` you'll see the following:

```
==========================
Interpolated string
==========================

val stringLiteral = s"$a"
---

(compilation_unit
  (val_definition (identifier)
    (string_transform_expression (identifier)
      (string (interpolation (identifier))))))
```

So this is correct if `string_transform_expression` is indeed the term we want to go for, but I sort of feel that having the following may be clearer:

```
==========================
Interpolated string
==========================

val stringLiteral = s"$a"
---

(compilation_unit
  (val_definition (identifier)
    (processed_string_expression (identifier)
      (string (interpolation (identifier))))))
```

The processed language comes from this [SIP](https://docs.scala-lang.org/sips/string-interpolation.html), and you can see this referenced in the [string interpolation](https://docs.scala-lang.org/overviews/core/string-interpolation.html) section of the docs as well.

> In the above, the literal s"Hello, $name" is a processed string literal. This means that the compiler does some additional work to this literal. A processed string literal is denoted by a set of characters preceding the ".

ckipp01:
This pr changes and enhances a bit of the way that string interpolation in modeled. Closes #14 

#### Naming change
`string_transform_expression` -> `interpolated_string_expression`

#### Behavioral changes

Previously interpolated strings were anything that contained a `$(identifier)` or `${identifier}` where now it _must_ be a `seq` of `identifier` and followed by a string.

Previously in `interpolate_string_expression`s the string was just a `string`. We now make a small distinction and the `string` inside of the `interpolated_string_expression` is a `interpolated_string` just to bring a bit more clarity.


corpus/literals.txt
```diff
@@ -1,43 +1,58 @@
...
 
 ==========================
 Interpolated strings
 ==========================
 
-val string1 = "a $b c ${d} e"
+val string1 = s"a $b ${c}"
 
-val string2 = "$a"
+val string2 = f"hi $name%s"
```
ckipp01, line 25: One future thing we can enhance even further here is identifying the formatters as identifiers, but for now there are a ton of other things that are probably more important to tackle than fine graining this.

maxbrunsfeld:
Awesome!

PR #22, opened 2021-03-21
Add in support for boolean, character, and symbol literals.

ckipp01:
Previously we would just handle boolean literals and character literals
as identifiers, and wouldn't handle symbol literals.


ckipp01:
Just realized I also need to support symbol literals as well.. I might as well add that in here too, so I moved to a draft for a bit.

grammar.js
```diff
@@ -640,6 +646,27 @@ module.exports = grammar({
...
+        seq('\\', choice(
+          /[^xu]/,
+          /u[0-9a-fA-F]{4}/,
+          /u{[0-9a-fA-F]+}/,
+          /x[0-9a-fA-F]{2}/
+        )),
+        /[^\\'\n]/
+      )),
+      '\''
+    )),
```
ckipp01, line 663: I went back and forth on this a bunch trying to exactly get what is described [here](https://www.scala-lang.org/files/archive/spec/2.13/01-lexical-syntax.html#character-literals), and finally I just went with a modified version of what tree-sitter-rust was using.

maxbrunsfeld:
Thanks! Sorry for the delay.

PR #25, opened 2021-06-23
Feature/literals

### LINKED ISSUE #24: Long literals parsed as infix_expressions instead of numbers
Hi tree-sitter-scala,

When parsing the following code block:
```
object Example {
  val element1 = 12345
  val element2 = 1234
}
```
Both `val element1 = 12345` and `val element2 = 1234` are parsed as `val_definition` nodes with the identifiers `element1`/`element2` and `12345`/`1234` parsed as numbers.

However if I change the code to:
```
object Example {
  val element1 = 12345L
  val element2 = 1234
}
```
Then a `val_definition` node is created for all of the code `val element1 = 12345L\nval element2 = 1234`, where the identifier is `element1` and `12345L\nval element2 = 1234` is treated as an `infix_expression` node.

gemelen:
Introduce hierarchy of literals as it is done in Scala language grammar.
Move existing literal types into hierarcy under `literal` parent.
Add and extend literals for integer and floating point numbers.
Add test cases for integer and floating point numbers in corpus.
Adjust existing test cases to changes.

Also fixes #24


corpus/annotations.txt
```diff
@@ -8,7 +8,7 @@ Classes
 
 (compilation_unit
   (class_definition
-    (annotation (type_identifier) (arguments (string) (string)))
+    (annotation (type_identifier) (arguments (literal (string)) (literal (string))))
```
maxbrunsfeld, line 11: Let's not add this wrapper `literal` node; we generally try to avoid having these single-child wrapper nodes in order to keep the tree concise and maximally efficient to traverse.

We can still keep the rule in the grammar by either by renaming it to `_literal` so that it's "hidden", or adding `$.literal` to the grammar's `inline` array. I would prefer whichever solution gives us the lower `STATE_COUNT` in the generated parser.

grammar.js
```diff
@@ -644,11 +636,61 @@ module.exports = grammar({
...
+            /[dfDF]/
+          )
+        )
+      )
+    ),
 
     boolean_literal: $ => choice('true', 'false'),
 
-   character_literal: $ => token(seq(
+    character_literal: $ => token(seq(
```
maxbrunsfeld, line 693: It looks like there's a stray whitespace added at the start of this line.

maxbrunsfeld (changes requested):
Thanks for adding this! I left two suggested changes.

grammar.js
gemelen: This one is intentional to align indentation of `character_literal` to its neighbours.

corpus/annotations.txt
gemelen: Both approaches return the same zero-value difference for `STATE_COUNT`.
I'll stick with name without an underscore and using `inline` array.

gemelen:
It seems that CI test doesn't include parser rebuild?
Should I then commit files related to parser itself that are generated by `npm run build` (this is not mentioned in readme)?

maxbrunsfeld:
Yeah, commit changes to any files that are checked in (`parser.c`, `grammar.json`, `node-types.json`).

maxbrunsfeld (approved):

maxbrunsfeld:
Thanks for the fix, and for writing good tests!

PR #26, opened 2021-07-01
Add postfix expression to grammar

gemelen:
Add precedence level for postfix operations.
Change ordinals of other levels of higher precendences.
Add grammar construct for postfix operations.
Add test cases.


gemelen:
¯\_(ツ)_/¯

PR #27, opened 2021-07-03
Miscallleneous stuff

gemelen:
Few changes to cover small gaps.
Most important is `unit` literal, required for `for-expression` even in naive implementation.


gemelen:
@maxbrunsfeld could you please have a look? (and also to #26)

grammar.js
```diff
@@ -751,6 +755,14 @@ module.exports = grammar({
       $._automatic_semicolon
     ),
 
+    null_literal: $ => token('null'),
```
maxbrunsfeld, line 758: I don't think you need the `token()` call around the string literal.

That function is only needed when you're using other rule-builder functions like `seq`/`repeat`/`choice`, and you want to tell Tree-sitter that the whole rule should be a single token. With a string literal, it's already a single token by default.

maxbrunsfeld (commented):
This looks good. I just left one minor suggestion.

maxbrunsfeld:
Thanks!

Issue #28, opened 2021-07-20
highlight groups


stumash:
The [list of languages with tree-sitter highlighting](https://github.com/nvim-treesitter/nvim-treesitter#supported-languages) doesn't include scala, because tree-sitter-scala is 'missing' a `queries/highlights.scm` file [like java has](https://github.com/tree-sitter/tree-sitter-java/blob/master/queries/highlights.scm).

There must be a vim user among **ye beneficent contributors**?

ckipp01:
@stumash keep in mind that for `nvim-treesitter` to use the queries they need to be defined in https://github.com/nvim-treesitter/nvim-treesitter/tree/master/queries, not really in this repo. I know some tree sitter repos do have them defined in their language repo in order to power highlighting for other tools etc, but `nvim-treesitter` specifically pulls them from their own queries.

stumash:
Hmm I didn't know that, thanks. Going by the fact that the `highlights.scm` file for java in the tree-sitter main repo is much bigger than the one in the tree-sitter-java repo, I think my changes should go in that repo, not this one.

I've got some reading to do.

PR #29, opened 2021-07-21
start the highlights.scm file

stumash:
got it started as per https://github.com/tree-sitter/tree-sitter-scala/issues/28

no testing but insanely basic

used this for the keyword list https://scala-lang.org/files/archive/spec/2.13/01-lexical-syntax.html
(removed `true` and `false` since they're technically `reserved words` according to the spec but should be highlighted as literals, not keywords)


stumash:
I'm closing this because of the conversation that was had over in #28

Issue #21, opened 2021-03-12
Unhandled block expression as parameter


ckipp01:
In Scala when you have a method that expects a single parameter the parens may be replaced with curly braces. We currently don't account for this.

For example:

```scala
object Main {
  def x(a: Int) = a

  x { 3 }
  x(3)
}
```

While a bit contrived, both of these are valid. However, currently the above will return:

```
(compilation_unit [0, 0] - [6, 0]
  (object_definition [0, 0] - [5, 1]
    name: (identifier [0, 7] - [0, 11])
    body: (template_body [0, 12] - [5, 1]
      (function_definition [1, 2] - [1, 19]
        name: (identifier [1, 6] - [1, 7])
        parameters: (parameters [1, 7] - [1, 15]
          (parameter [1, 8] - [1, 14]
            name: (identifier [1, 8] - [1, 9])
            type: (type_identifier [1, 11] - [1, 14])))
        body: (identifier [1, 18] - [1, 19]))
      (ERROR [3, 2] - [3, 3]
        (identifier [3, 2] - [3, 3]))
      (block [3, 4] - [3, 9]
        (number [3, 6] - [3, 7]))
      (call_expression [4, 2] - [4, 6]
        function: (identifier [4, 2] - [4, 3])
        arguments: (arguments [4, 3] - [4, 6]
          (number [4, 4] - [4, 5]))))))
```

The `ERROR` here should still be a `call_expression`. ~I'm a little unsure what to call the `{ 3 }` since `arguments` is valid, but also a bit wrong since there can only be one. So I'm trying to think if this should be like `single_block_argument` or something like that signifying that it is a block, but an argument as well. I also can't really find reference to this in the spec to see exactly how this should be called.~ So I was a bit wrong on this, it's just a multiline block expression, but on a single line.

stumash:
ya, I often write code like this:

```scala
List(1,2,3).map { x =>
  f(x) // maybe do slightly more complicated things than f(x)
}
```

I think the gramar should be able to parse this as (rougly)
```scheme
(call_expression
  function: (identifier)
  arguments: (anonymous_function_expression))
```

I'm working on a PR, but I'm getting a gnarly amount of conflicts from adding `anonymous_function_expression`.
`(x, y) => e` looks like a tuple until the `=>`. I'm really putting the `G` in `GLR` parser 😢

stumash:
[...]

PR #34, opened 2021-09-01
Add lambda expression and fix call expression

stevanmilic:
Add lambda expression

Fix call expression to allow block as arguments (#21)


grammar.js
```diff
@@ -498,7 +502,7 @@ module.exports = grammar({
     tuple_pattern: $ => seq(
       '(',
       $._pattern,
-      repeat1(seq(',', $._pattern)),
+      repeat(seq(',', $._pattern)),
```
stumash, line 505: 👍

grammar.js
```diff
@@ -560,11 +565,33 @@ module.exports = grammar({
...
+    bindings: $ => seq(
+        '(',
+        commaSep($.binding),
+        ')',
+    ),
+
+    lambda_expression: $ => prec.right(PREC.lambda, seq(
+        choice(
+            $.bindings,
+            $.identifier,
```
stumash, line 582: why does it have to be a choice of `binding` or `identifier` if `binding` has a `name: (identifier)`?

grammar.js
```diff
@@ -807,16 +833,13 @@ module.exports = grammar({
...
-      )),
+      sep1($._semicolon, $.enumerator),
       optional($._automatic_semicolon)
     ),
 
     enumerator: $ => seq(
-      $.identifier,
-      '<-',
+      $._pattern,
+      choice('<-', '='),
```
stumash, line 842: apologies for the dumb question. do you mind spelling out for me a scala example of what this change achieves

grammar.js
```diff
@@ -807,16 +833,13 @@ module.exports = grammar({
...
-        $.assignment_expression
-      )),
+      sep1($._semicolon, $.enumerator),
       optional($._automatic_semicolon)
     ),
 
     enumerator: $ => seq(
-      $.identifier,
-      '<-',
+      $._pattern,
```
stumash, line 841: while we're at it, should `$.assignment_expression` itself be using `$._pattern` for its LHS too?

grammar.js
stevanmilic: it's a choice of `bindings` (mind the `s`) and `identifier`. And the reason why we use the `identifier` is because of the `a => a + 1` use-case, where you can't have a type for it i.e. `a: Int => a + 1`.

grammar.js
```diff
@@ -591,8 +618,7 @@ module.exports = grammar({
 
     call_expression: $ => prec(PREC.call, seq(
       field('function', $.expression),
-      field('arguments', $.arguments),
-      field('body', optional(choice($.block, $.case_block)))
+      field('arguments', choice($.arguments, $._case_clauses, $.block)),
```
stumash, line 621: I get why this is necessary, but I was definitely hoping that _somehow_ we could get the `$._case_clauses` in

```scala
LIst(1,2,3).map {
  case x if x < 2 => "yes"
  case x => "no"
}
```

to be recognized as a `$.lambda_expression`

corpus/expressions.txt
```diff
@@ -529,8 +583,9 @@ def f() = {
      (for_expression
        (enumerators
          (enumerator (identifier) (identifier))
-         (assignment_expression (identifier) (infix_expression (identifier) (operator_identifier) (identifier)))
-         (enumerator (identifier) (call_expression (identifier) (arguments)) (guard (identifier))))
+         (enumerator (identifier) (infix_expression (identifier) (operator_identifier) (identifier)))
+         (enumerator (identifier) (call_expression (identifier) (arguments)) (guard (identifier)))
+        (enumerator (tuple_pattern (identifier) (identifier)) (identifier)))
```
stumash, line 588: pedantry: indentation

grammar.js
stevanmilic: hmm but that's not a lambda expression, it's essentially a block expression. From the Scala syntax summary:

```
  ArgumentExprs     ::=  ‘(’ [Exprs] ‘)’
                      |  ‘(’ [Exprs ‘,’] PostfixExpr ‘:’ ‘_’ ‘*’ ‘)’
                      |  [nl] BlockExpr
  BlockExpr         ::=  ‘{’ CaseClauses ‘}’
                      |  ‘{’ Block ‘}’```

File: grammar.js (hunk 2 above)
Comment (stumash) [reply grammar.js:582]: okay, so correct me if I'm wrong, but I always thought that these parsers don't actually have the same goal as regular parsers, which have to both
1. reject incorrect code
2. accept correct code

to my eye, tree-sitter parsers actually have different goals, that is
1. accept correct code
2. make predictable ASTs
3. handle bad code gracefully so that the rest of the AST doesn't get ruined

I'm not trying to be pushy, and it's definitely a judgement call as to whether you leave it the way that it is, but in my mind it's preferable that **all `lambda_expression`s have a `bindings` node** than that incorrect lambda expressions are rejected. Of course, if allowing the incorrect `: TypeName` causes conflicts, then I understand. But otherwise, I would actually prefer that ALL `lambda_expressions` have a predictable `bindings` node.

File: grammar.js (hunk 5 above)
Comment (stumash) [reply grammar.js:621]: you're right. it's semantically an anonymous function, but lexically not! okay then.

File: corpus/expressions.txt [hunk 7]
```diff
@@ -216,6 +231,7 @@ def matchTest(x: Int): String = x match {
   case 3 => {
     "3"
   }
+  case ((i, _)) => i
```
stumash, line 234: how come you're nesting the tuple pattern? `case (x, y) => z` is acceptable, you don't need `case ((x, y)) => z`

corpus/expressions.txt
```diff
@@ -7,6 +7,11 @@ class C {
     a()
     a(123)
     a(b, c)
+    a.map(x => x + 1)
+    a { 123 }
+    a { b }
+    a.map{ x => x + 1 }
+    a.exists { case ((i, _)) => i }
```
stumash, line 14: same here. how come nested tuple?

grammar.js
stumash: It' s a matter of preference, and if you reply "meh I prefer it my way" I'm not gonna push this any further

stumash:
anyway, other than the things I brought, LGTM!

corpus/expressions.txt
stevanmilic: fair point, I was just testing that particular use-case, since it's also an acceptable syntax, and I had to make some modifications in order for it to work.

corpus/expressions.txt
stevanmilic: https://github.com/tree-sitter/tree-sitter-scala/pull/34#discussion_r702392299

grammar.js
stevanmilic: > I always thought that these parsers don't actually have the same goal as regular parsers

Fair point, I was also thinking about it. My main concern is how the AST will be worked upon, once the tree-sitter does it job e.g. when you use nvim-tree-sitter-text-objects you would want your expression to be exactly what they are in order to make a change accordingly. OTOH handling bad code gracefully is something that's definitely required, in this particular case tho it would parse `a: Int => a + 1` as an `ascription_expression`  and an `infix_expression` – so it won't break. I think it's fine to stay as-is :)

stevanmilic:
@maxbrunsfeld can you take a look at this PR? and also https://github.com/tree-sitter/tree-sitter-scala/pull/35?

grammar.js
```diff
@@ -560,11 +580,33 @@ module.exports = grammar({
...
 
+    binding: $ => prec.dynamic(PREC.binding, seq(
+        field('name', $.identifier),
+        optional(seq(':', field('type', $._param_type))),
+    )),
+
+    bindings: $ => seq(
+        '(',
+        commaSep($.binding),
+        ')',
```
maxbrunsfeld, line 591: Minor nitpick - Could you make the indentation for the new lines consistent with the rest of the file?

grammar.js
```diff
@@ -560,11 +580,33 @@ module.exports = grammar({
...
+        $.expression,
+    )),
+
     case_block: $ => choice(
       prec(-1, seq('{', '}')),
-      seq('{', repeat1($.case_clause), '}')
+      $._case_clauses,
     ),
 
+    _case_clauses: $ => seq('{', repeat1($.case_clause), '}'),
```
maxbrunsfeld, line 608: Just curious - what was the purpose of extracting this rule?

maxbrunsfeld (commented):
Looks good! Thanks for adding good tests. I just left a couple of minor questions.

grammar.js
stevanmilic: Hmm good catch, it's not needed. I remember adding it because it's used in the `call_expression`, to dissalow `f{}`, but that's a valid expression.

PR #35, opened 2021-09-02
call_expression block argument

stumash:
I decided to save myself the out-of-scope trouble and not implement `anonymous_function_expression`.

I'll do that next.

Like I said [here](https://github.com/tree-sitter/tree-sitter-scala/issues/21#issuecomment-898645991), it's gonna cause a ton of conflicts with `tuple`s.


stumash:
the reason I created the `class_arguments` node is because in scala

```scala
class Foo(f: Int => Int ) {
  def doIt(x: Int): Int = f(x)
}

// you're not allowed to to just do `new Foo { x => x + 1 }`
new Foo({ x => x + 1 })
  .doIt(3)


// but you CAN for FUNCTION calls
def someFunc(f: Int => Int): Int = f(3)
someFunc { x => x + 1 }
```

and it caused conflicts because

```scala
class Foo(f: Int => Int ) {
  def doIt(x: Int): Int = f(x)
}

// is this `{ ... }` the body of the anonymous class instance or the argument to the constructor?
new Foo { x => x + 1 }
```

stumash:
closing because idle for long time

ckipp01:
Btw I know you pinged me on this @stumash, but I have no merge rights or anything on this repo. I believe @maxbrunsfeld is the only one with those.

PR #46, opened 2022-02-02
Handle Scala 3 macro expressions

keynmol:
At this third PR I'm starting to think I should've merged them all in one because of PR conflicts :-/

- This adds very basic support for `'{` and '${' quote/splice expressions
- Also fortify symbol literal definition to both avoid ambiguity with `'{` but also to follow the spec

### DIFF
--- a/corpus/scala3.txt
+++ b/corpus/scala3.txt
@@ -1,0 +1,23 @@
+===================================
+Macro expressions
+===================================
+
+def string(arg: Int) = 
+  ${ '{ ${ yesh('arg) } } }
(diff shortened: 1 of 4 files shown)


stumash:
is this project supposed to support scala 3 / dotty syntax in the first place? aren't there some major changes to syntax?

sideeffffect:
@keynmol would you like to reopen the PR in light of https://github.com/tree-sitter/tree-sitter-scala/issues/63#issuecomment-1369842151 ?

PR #48, opened 2022-03-27
Add syntax highlighting queries

Wilfred:
This is heavily based on tree-sitter-java, but with grammar rules and
keyword names changed where appropriate.


aryx (approved):

aryx:
the CI error for windows is because we're still used nodejx 14 and we must switch to 16. The same happen in tree-sitter-java and tree-sitter-rust last week.

PR #61, opened 2022-12-14
Optional braces, part 1

eed3si9n:
Problem
-------
Currently Scala 3 optional braces syntax does not work.

Solution
--------
This implements initial attempt on dealing with colon instead of braces.
Note that this does not implement any indent/outdent tracking so the second `val` onwards will not be added in `template_body` but it should improve the highlighting a bit.

Demo
------
Here's a demonstration of before and after using Sonokai. The theme is a bit too color happy, but it's good for checking what changed:

## Before
sonokai_main_before

**Note**: `def run` is not recognized as a method.

## After
sonokai_main_after

### DIFF
--- a/corpus/definitions.txt
+++ b/corpus/definitions.txt
@@ -16,6 +16,39 @@ package c {
       (object_definition (identifier)))))
 
 ================================
+Package (Scala 3 syntax)
+================================
+
+package a.b
+package c:
+  object A
+
+---
+
+(compilation_unit
+  (package_clause (package_identifier (identifier) (identifier)))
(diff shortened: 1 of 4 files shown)


ckipp01:
@maxbrunsfeld sorry to ping you on here, but I'm unsure how else to reach you. Looking at the open and closed prs on this repo it's been a bit of a struggle to get reviews and get changes merged in. There's an uptick in usage right now from the Scala community on this repo, and I'd hope to not lose that.

Are you open to having more maintainers of the repo? I'd be happy to step up and play a larger role in reviewing and getting things across the board here if that's a possibility.

Issue #43, opened 2021-11-02
Scala 3 syntax support?


peterstorm:
Hello! 

Just wondering if Scala 3 syntax support was in the works?

Specifically the braceless syntax.

Thank you for your work!

eed3si9n:
Here's my PR for the optional braces support - #62

ckipp01:
Now that #62 has been merged in we'll start to focus on individual parts of the missing grammar. So in short, the answer to your question @peterstorm is yea, it's in the works 👍🏼 .

Please do report issues as we're actively working on stuff.

PR #62, opened 2022-12-15
Optional braces, part 2

eed3si9n:
Ref https://github.com/tree-sitter/tree-sitter-scala/issues/43

Problem
-------
Currently Scala 3 optional braces syntax does not work.
Specifically, in this PR I want to address decls/defns not properly grouped under the right owner in part 1 (#61).

Solution
--------
1. This brings in indent/outdent impl by @keynmol
2. Introduce `_indentable_expression`.
   This allows us to limit where indent/outdent shows up without
   confusing tree-sitter too much.
   Internally I've made a change to track `last_indentation_size` in the payload to indicate last
    outdent (default: -1). If it's not -1 that means there was an outdent.
    This then checks if the last_indentation_size qualified for another outdent.
2. Add support for control strctures:
    - [x] if expression
    - [x] try-catch expression
    - [x] match expression
    - [x] while expression
    - [x] for expression    

The test cases for these are:

```lisp
=======================================
Top-level Definitions (Scala 3 syntax)
=======================================

class A:
  def a() =
    ()

def a() = 1

---

(compilation_unit
  (class_definition
    (identifier)
    (template_body
      (function_definition
        (identifier)
        (parameters)
        (indented_block
          (unit)))))
  (function_definition
    (identifier)
    (parameters)
    (integer_literal)))
```

```diff
  =======================================
  Value declarations (Scala 3 syntax)
  =======================================
  
  class A:
-   val b: String
+   val b, c : Int
+   val d : String
```

Demo
------
Here's a demonstration of before and after using Sonokai. The theme is a bit too color happy, but it's good for checking what changed:

## Before #61
sonokai_main_before

**Note**: `def run` is not recognized as a method.

## After this PR

sonokai_main_after2

`def run`  is highlighted as a method. The body of `def dealiasBaseDirectory` improved as well, since `val` is now acceptable without `{ ... }`.

### DIFF
--- a/src/stack.h
+++ b/src/stack.h
@@ -16,6 +16,8 @@ typedef struct ScannerStack {
   unsigned int stack[STACK_SIZE];
   int top;
   int last_indentation_size;
+  int last_newline_count;
+  int last_column;
 } ScannerStack;
 
 ScannerStack* createStack() {
@@ -23,6 +25,8 @@ ScannerStack* createStack() {
 
   ptr -> top = 0;
   ptr -> last_indentation_size = -1;
+  ptr -> last_newline_count = 0;
+  ptr -> last_column = -1;
   memset(ptr -> stack, STACK_SIZE, (0));
 
   return ptr;
@@ -60,10 +64,12 @@ void printStack(ScannerStack *stack, char *msg) {
 
 unsigned serialiseStack(ScannerStack *stack, char *buf) {
   unsigned elements = isEmptyStack(stack) ? 0 : stack->top;
-  unsigned result_length = (elements + 1) * sizeof(int);
+  unsigned result_length = (elements + 3) * sizeof(int);
   int *placement = (int *)buf;
   memcpy(placement, stack->stack, elements * sizeof(int));
   placement[elements] = stack->last_indentation_size;
+  placement[elements + 1] = stack->last_newline_count;
+  placement[elements + 2] = stack->last_column;
 
   return result_length;
 }
@@ -72,14 +78,18 @@ void deserialiseStack(ScannerStack* stack, const char* buf, unsigned n) {
   if (n != 0) {
     int *intBuf = (int *)buf;
 
-    unsigned elements = n / sizeof(int) - 1;
+    unsigned elements = n / sizeof(int) - 3;
     stack->top = elements;
     memcpy(stack->stack, intBuf, elements * sizeof(int));
     stack->last_indentation_size = intBuf[elements];
+    stack->last_newline_count = intBuf[elements + 1];
+    stack->last_column = intBuf[elements + 2];
   }
 }
 
 void resetStack(ScannerStack *p) {
   p->top = 0;
   p->last_indentation_size = -1;
+  p->last_newline_count = 0;
+  p->last_column = -1;
 }
--- a/src/stack.h
+++ b/src/stack.h
@@ -1,0 +1,78 @@
+
+#include <stdio.h>
+#include <stdbool.h>
+#include <stdlib.h>
+#include <string.h>
+
+#ifdef DEBUG
+    #define LOG(args...) fprintf(stderr, args);
+#else
+    #define LOG(args...)
+#endif
+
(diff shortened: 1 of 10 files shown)


bindings/rust/build.rs
```diff
@@ -5,7 +5,9 @@ fn main() {
     c_config
         .flag_if_supported("-Wno-unused-parameter")
         .flag_if_supported("-Wno-unused-but-set-variable")
-        .flag_if_supported("-Wno-trigraphs");
+        .flag_if_supported("-Wno-trigraphs")
+        .flag_if_supported("-Werror")
+        .flag_if_supported("-W!all"); // -Wextra -Wpedantic -Wvla -Wconversion
```
keynmol, line 10:
```suggestion
        .flag_if_supported("-W!all"); // -Wextra -Wpedantic -Wvla -Wconversion
```

This was me testing if this is where the flags are passed to clang. It didn't work, this line can be removed

src/stack.h
```diff
@@ -0,0 +1,81 @@
...
+#include <string.h>
+
+#ifdef DEBUG
+    #define LOG(args...) fprintf(stderr, args);
+#else
+    #define LOG(args...)
+#endif
+
+typedef struct ScannerStack {
+  unsigned int stack[256];
```
keynmol, line 14: This stack does not re-allocate if it reaches the limit and I've chosen a rather optimistic 256 levels of indentation as being absolute maximum that a sane person would use..

src/stack.h
```diff
@@ -0,0 +1,81 @@
...
+  }
+  LOG("]\n");
+}
+
+unsigned serialiseStack(ScannerStack *stack, char *buf) {
+  unsigned elements = isEmptyStack(stack) ? 0 : stack->top;
+  unsigned result_length = elements * sizeof(int);
+  int *placement = (int *)buf;
+
+  for (int i = 0; i < elements; i++) {
```
keynmol, line 62: tbf this could be `memcpy` and much faster, but I don't have intuition on how often tree-sitter serialises the stack during a normal parsing

eed3si9n: Sounds like a good idea.

bindings/rust/build.rs
eed3si9n: Addressed in 0002b5f09453f48b80405020fb76db3d6c9088e3

src/stack.h
eed3si9n: Addressed in 0002b5f09453f48b80405020fb76db3d6c9088e3

ckipp01 (approved):
I don't have much to add apart from what @keynmol already mentioned. I think we're all on the same page that to fully handle the new syntax would need some more C work, but this is a significant improvement over what we had before.

Thanks @eed3si9n

keynmol:
I think before that we'll need to run this on a larger codebases - I.e. I've been using your fork and it still has problems in quite a few places, so I would suggest ignoring fewerBraces completely and instead getting the number of correctly parsed files up on something like Dotty codebase.

Additionally, would be good to keep track of Scala 2 as well, for which we can also use the code of Scala 2 compiler as a bench.

To get that locally, you can use this command (thanks to @ckipp01 for showing me the light):

```
> tree-sitter parse ~/projects/dotty/compiler/**/*.scala --quiet --stat | tail -n2
Total parses: 745; successful parses: 282; failed parses: 463; success percentage: 37.85%
```

ckipp01:
> I think before that (=fewer braces) we'll need to run this on a larger codebases - I.e. I've been using your fork and it still has problems in quite a few places, so I would suggest ignoring fewerBraces completely and instead getting the number of correctly parsed files up on something like Dotty codebase.

Yea I'm in agreement here. Even without fewer braces I still hit on a lot of issues. I think if we get your stuff revived and in @keynmol I can start paying way more attention to errors in the tree and reporting them now that I feel more confident that we can get things merged in.

eed3si9n:
Yea. Generally I agree that we should try to go after lower hanging Scala 3 constructs first, but I was replying to the "need some more C work" part.

PR #51, opened 2022-06-07
Handle backquoted identifiers

keynmol:
Admittedly, this doesn't make the identifier grammar any more complete (according to EBNF), but it fixes the "infinite recursion of doom"

image

(To see it for yourself, open this file: https://github.com/VirtusLab/scala-cli/blob/main/build.sc)

With this patch, all is great in the world again: 

image

And the backquoted identifiers are highlighted beautifully.

PR #65, opened 2023-01-03
Backquoted identifiers

keynmol:
Copy of https://github.com/tree-sitter/tree-sitter-scala/pull/51 but I force-pushed the branch on a closed PR and can no longer reopen it.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -68,7 +68,7 @@ module.exports = grammar({
     [$._indentable_expression, $.do_while_expression],
   ],
 
-  word: $ => $.identifier,
+  word: $ => $._plainid,
 
   rules: {
     compilation_unit: $ => repeat($._top_level_definition),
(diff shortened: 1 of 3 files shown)


grammar.js
```diff
@@ -778,7 +778,10 @@ module.exports = grammar({
     ),
 
     // TODO: Include operators.
-    identifier: $ => /[a-zA-Z_]\w*/,
+    _plainid: $ => /[a-zA-Z_]\w*/,
+    // inside ` corresponds roughly to charNoBackQuoteOrNewline in Scala 3 EBNF
+    _char_no_back_quote_or_newline: $=> /[^\n`]+/,
+    identifier: $ => choice($._plainid, seq('`', $._char_no_back_quote_or_newline, '`')),
```
eed3si9n, line 784: If you use `seq(...)` wouldn't the current grammar allow newline in between the backquote and `$._char_no_back_quote_or_newline`? i.e.

```scala
val `

test-this` = 1
```

keynmol: You are right, I changed it from a regex, and it is indeed incorrect.
Using token.immediate also didn't work (and I didn't have the patience to experiment given that the grammar takes 2 minutes at least to regenerate)

It's back to original form in https://github.com/tree-sitter/tree-sitter-scala/pull/65/commits/0d94e6ee2a3ce4494e3edbf8ce70bac7e6e3ef71

ckipp01 (approved):
👍🏼 Thanks @keynmol!

PR #68, opened 2023-01-04
Adds support for $ in identifiers.

nrfulton:


eed3si9n (commented):
@nrfulton Thanks for the contribution!

Overall LGTM. Could you add a corpus test around here please?
https://github.com/tree-sitter/tree-sitter-scala/blob/master/corpus/definitions.txt#L331-L333

nrfulton:
Thanks, @eed3si9n. I added a corpus test for the change.

eed3si9n (approved):
Thanks!

Issue #49, opened 2022-04-10
Release and tag a new version


jirutka:
Can you please release a new version? The latest one (v0.19.0) doesn’t work with modern nvim-treesitter.

I’m maintainer of the [tree-sitter-scala](https://pkgs.alpinelinux.org/packages?name=tree-sitter-scala) package in Alpine Linux and I had to update the package to the latest revision in the master branch, instead of using the tagged release, which is not optimal.

Will-Sommers:
This would be super helpful for us too!

eed3si9n:
@ckipp01 Should we bump the version number in package.json and add a new tag for 0.19.1?

ckipp01:
> Can you please release a new version? The latest one (v0.19.0) doesn’t work with modern nvim-treesitter.

Hey @jirutka just to make sure I understand the actual issue, could you expand a bit more on this? When using `nvim-treesitter` it should actual pull the latest revision [in the lock file](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/lockfile.json#L336). Where does the version actually come into play and cause an issue? Just asking to better understand myself.

> @ckipp01 Should we bump the version number in package.json and add a new tag for 0.19.1?

Sure!

PR #69, opened 2023-01-05
Bump to 0.19.1

eed3si9n:
Fixes https://github.com/tree-sitter/tree-sitter-scala/issues/49

Once this is merged, I'll tag the commit.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "tree-sitter-scala"
 description = "scala grammar for the tree-sitter parsing library"
-version = "0.19.0"
+version = "0.19.1"
 keywords = ["incremental", "parsing", "scala"]
 categories = ["parsing", "text-editors"]
 repository = "https://github.com/tree-sitter/tree-sitter-javascript"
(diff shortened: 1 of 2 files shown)


ckipp01 (commented):
👍🏼 we can totally do this. I don't see any issue with it. However I think we should also bump the [Cargo.toml](https://github.com/tree-sitter/tree-sitter-scala/blob/master/Cargo.toml#L4) at the same time right?

I was curious and looking around at the other grammars, it seems that bumping the version is actually pretty rare.

- Last bump in the go one was [over a year ago](https://github.com/tree-sitter/tree-sitter-go/commit/e41dd569d91eb58725baa7089c34fc3d785b2978)
- [Same with cpp](https://github.com/tree-sitter/tree-sitter-cpp/commit/f44509141e7e483323d2ec178f2d2e6c0fc041c1)

eed3si9n:
> 👍🏼 we can totally do this. I don't see any issue with it. However I think we should also bump the [Cargo.toml](https://github.com/tree-sitter/tree-sitter-scala/blob/master/Cargo.toml#L4) at the same time right?

Done

PR #81, opened 2023-01-05
Smoke test

eed3si9n:
This adds real-world integration test using scala/scala and Dotty compiler code base.
The expected parse success% is currently hardcoded, but at least we'll catch if this number degrades.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,8 +11,27 @@ jobs:
       fail-fast: true
       matrix:
         os: [macos-latest, ubuntu-20.04, windows-2019]
+    env:
+      SCALA_SCALA_DIR: /home/runner/work/tree-sitter-scala/tree-sitter-scala/scala_scala
+      DOTTY_DIR: /home/runner/work/tree-sitter-scala/tree-sitter-scala/dotty
     steps:
-      - uses: actions/checkout@v3
+      - name: checkout tree-sitter-scala
+        uses: actions/checkout@v3
+
+      - name: checkout scala/scala
+        uses: actions/checkout@v3
+        with:
+          repository: scala/scala
+          ref: v2.13.10
+          path: scala_scala
+
+      - name: checkout lampepfl/dotty
+        uses: actions/checkout@v3
+        with:
+          repository: lampepfl/dotty
+          ref: 3.2.1
+          path: dotty
+
       - uses: actions/setup-node@v3
         with:
           node-version: 18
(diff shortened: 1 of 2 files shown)


eed3si9n:
Looks like this:
Smoke_test

script/smoke_test.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash -e
+
+# This is an integration test to generally check the quality of parsing.
+
+SCALA_SCALA_LIBRARY_EXPECTED=87
+SCALA_SCALA_COMPILER_EXPECTED=46
+DOTTY_COMPILER_EXPECTED=35
```
keynmol, line 7: _sad trombone music_

eed3si9n: Only way from here is up. *optimistic trombone music*

SethTisue: https://orchestraexcerpts.com/trombone-wagner-die-walkure-ride-valkyries-2-excerpts/

script/smoke_test.sh
```diff
@@ -0,0 +1,45 @@
...
+
+run_tree_sitter () {
+  local source_dir=$1
+  local expected=$2
+  echo "Parse $source_dir:"
+  out=$(tree-sitter parse "$source_dir"/**/*.scala --quiet --stat || true)
+  actual=$(echo $out | rev | cut -d' ' -f1 | rev | sed 's/.$//')
+  echo "$out"
+  if (( $(echo "$actual > $expected" |bc -l) )); then
+    echo -e "  - ${Green}${actual}% ok${Color_Off}"
```
keynmol, line 32: Absolutely unnecessary, but would be nice to have: we could emit this green/red warning as a Github PR annotation: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#example-creating-an-annotation-for-an-error

Basically doing

```bash
echo "::error file=grammar.js,line=1::Missing semicolon"
```

Will show a message on the PR (similar to how coverage tools do it).

.github/workflows/ci.yml
```diff
@@ -26,6 +45,13 @@ jobs:
           npm install && npm run build
           git diff --quiet || (echo "::error file=grammar.js::Generated parser differs from the checked in version" && exit 1)
 
-      - name: Parser tests 
+      - name: Parser tests
         shell: bash
         run: npm install && npm run test
+
+      - name: Smoke test
+        if: ${{ matrix.os == 'ubuntu-latest' }}
```
keynmol, line 53:
```suggestion
        if: ${{ matrix.os == 'ubuntu-20.04' }}
```

Now that I think about it's probably easier to use `runner.os == "Linux"` if that's the right value https://docs.github.com/en/actions/learn-github-actions/environment-variables#detecting-the-operating-system

.github/workflows/ci.yml
```diff
@@ -26,6 +45,13 @@ jobs:
...
+        shell: bash
+        run: |
+          export PATH="$PATH:node_modules/.bin/"
```
keynmol, line 56: I believe you can avoid this and use `npm exec`, i.e.

```bash
npm exec -c 'tree-sitter --help'
```

eed3si9n: Yea. Let me try that.

script/smoke_test.sh
eed3si9n: Done

keynmol: Nice, I can see the correct messages in the actions log.

Your #82 will be the first one to demonstrate this working, as this PR doesn't touch grammar.js file so I guess the message just gets ignored.

eed3si9n:
I've addressed @keynmol's reviews, so I am going to self-merge.

Issue #71, opened 2023-01-05
Using clauses


eed3si9n:
https://docs.scala-lang.org/scala3/reference/contextual/using-clauses.html

In Scala 3, the implicit parameters use `using` keyword:

```scala
def max[T](x: T, y: T)(using ord: Ord[T]): T =
  if ord.compare(x, y) < 0 then y else x
```

eed3si9n:
Here's my PR for this - https://github.com/tree-sitter/tree-sitter-scala/pull/82

PR #82, opened 2023-01-05
Using clauses

eed3si9n:
Fixes https://github.com/tree-sitter/tree-sitter-scala/issues/71

### Problem
Scala 3 syntax uses `using` to denote implicit parameters.

### Solution
Allow both `implicit` and `using`.

### Note
This improves Dotty parsing success% (tested using https://github.com/tree-sitter/tree-sitter-scala/pull/81) from 35.68% to 56.81%

### DIFF
--- a/corpus/definitions.txt
+++ b/corpus/definitions.txt
@@ -263,6 +263,23 @@ class Point(val x: Int, val y: Int)(implicit coord: Coord)
       (class_parameter (identifier) (type_identifier)))))
 
 =======================================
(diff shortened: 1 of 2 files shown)


ckipp01 (approved):
Simple enough. LGTM!

> This improves Dotty parsing success% (tested using https://github.com/tree-sitter/tree-sitter-scala/pull/81) from 35.68% to 56.81%

Ha, noice.

PR #83, opened 2023-01-06
Generate and check parser code on CI

keynmol:
This is to ensure that whatever `parser.c` is added to the PR is indeed generated from the grammar (because reviewing the generated code is not possible)


.github/workflows/ci.yml
```diff
@@ -10,15 +10,22 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        os: [macos-latest, ubuntu-latest, windows-2019]
+        os: [macos-latest, ubuntu-20.04, windows-2019]
```
keynmol, line 13: Drop down to 20.04 due to https://github.com/actions/runner-images/issues/6709

eed3si9n (approved):
Thanks, @keynmol!

PR #86, opened 2023-01-06
Generate code on non-Linux

eed3si9n:
Problem
-------
Currently code is generated only for Linux.
Also the parser check prevents us from omitting codegen in PR.

Solution
--------
1. generate parser for non-Linux as well
2. run the parser check only if the PR contains parser.c


keynmol:
Ugh, MacOS is upwards of 11 minutes now :( 

This makes none of the original idea feasible.

eed3si9n:
Going to self-merge to unblock open PRs.

Issue #85, opened 2023-01-06
Improve contribution and PR process


keynmol:
Aim of this is to increase contribution velocity on this repository without impacting users.

### Current situation

1. If you have 2 PRs changing any two parts of the grammar, they will immediately conflict with each other because of the dramatic changes in generated code (usually those changes are not mergeable)
2. This means parallel PRs are in constant queue and need to be updated after each merge to the master

### Suggestion

1. Grammar PRs _only_ bring in changes to `grammar.js`, `corpus/*` or `src/scanner.c` - **not** `parser.c` or any other generated code
    - This will dramatically reduce the conflicts and will make all PRs more likely to be mergeable without manual updates
2. CI still checks for parser code being generated from grammar alone, and run tests against that
3. When we feel like a release is due, we actually generate parser code and commit it as a separate PR

Because of the CI checks, the above process should be safe, at the cost of not containing the latest-est generated code in `master`. Which IMO is not a big deal.

eed3si9n:
I like the nature that this repository is ready-to-use from the consumer's perspective, even though it is _kind of weird_ that there's a generated code committed inside. The "release" approach I think requires human attention going forward, so I am not sure if that's a sustainable approach. I don't want to rely too much on high energy.

Alternatively, given the "current situation" mostly happens between you and me for some span of time, maybe we can come up with a convention for ourselves?:

### Suggestion 2
1. Our grammar PRs _only_ bring in changes to `grammar.js`, `corpus/*` or `src/scanner.c` - **not** `parser.c` or any other generated code
    - This will dramatically reduce the conflicts and will make all PRs more likely to be mergeable without manual updates
2. CI still checks for parser code being generated from grammar alone, and run tests against that
3. Per PR, we immediately send a second PR with generated code inside, and merge it upon the grammar PR merge.
    - Since other grammar PRs do not conflict, this should only conflict the other codegen PRs only
4. Run a weekly cron job to check if the generated code is fresh

What do you think?

keynmol:
>     Since other grammar PRs do not conflict, this should only conflict the other codegen PRs only

I think this gets us back to the same situation - if each grammar PR has a corresponding codegen PR, then you have the same number of conflicts and same linear queue of PRs, each of them needs to be updated manually (even if update is literally `npm run build && git commit --amend --no-edit && git push -f`.

Plus I think it introduces more process to avoid forgetting to send the codegen PR for the right change.

[...]

PR #87, opened 2023-01-06
Opaque type aliases [grammar, highlights, tests]

### LINKED ISSUE #77: Opaque type alias
https://docs.scala-lang.org/scala3/reference/other-new-features/opaques.html

```scala
object MyMath:

  opaque type Logarithm = Double
```

keynmol:
Closes #77 

As discussed in #85, only adding non-generated code changes to avoid PR chain and conflicts


eed3si9n:
Re: non-Linux, this would need https://github.com/tree-sitter/tree-sitter-scala/pull/86 first.

ckipp01:
I'm not really sure why, but it seems this is bumping generation time on mac over 20 minutes...

PR #88, opened 2023-01-06
Literal types

### LINKED ISSUE #84: Literal types
```scala
type A = "hello"
type B = 25
type C = false
type D = 0.5f

class Test(d: "hello", b: 25)
```

Scala 2.13/3

Can be adapted from https://github.com/keynmol/tree-sitter-scala/pull/1/commits/145374be14be0a2a0478a732029f5cee372d3486

keynmol:
I had to separate null and non null because the former cannot be a literal type 
image

Closes #84

### DIFF
--- a/corpus/types.txt
+++ b/corpus/types.txt
@@ -250,3 +250,41 @@ type A = B with B1 with B2 ! C with C1
       (compound_type (type_identifier) (type_identifier) (type_identifier))
       (operator_identifier)
       (compound_type (type_identifier) (type_identifier)))))
+
+==================================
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Thanks @keynmol!

ckipp01:
Since https://github.com/tree-sitter/tree-sitter-scala/pull/91 fails since it doesn't regenerate the parser there I'm just going to piggy back on your PR here @keynmol. I went ahead and rebased, and then added another commit bumping the percentages.

Issue #72, opened 2023-01-05
Enumerations


eed3si9n:
https://docs.scala-lang.org/scala3/reference/enums/enums.html

```scala
enum Option[+T]:
  case Some(x: T) extends Option[T]
  case None       extends Option[Nothing]

enum Color(val rgb: Int):
  case Red   extends Color(0xFF0000)
  case Green extends Color(0x00FF00)
  case Blue  extends Color(0x0000FF)
  case Mix(mix: Int) extends Color(mix)
```

https://docs.scala-lang.org/scala3/reference/contextual/derivation.html

```scala
enum Tree[T] derives Eq, Ordering, Show:
  case Branch(left: Tree[T], right: Tree[T])
  case Leaf(elem: T)
```

keynmol:
Starting point can be copied from https://github.com/keynmol/tree-sitter-scala/pull/1/commits/df196cbec53717f7c1146156c63d90966e68f58e

keynmol:
I think with fixes in #117 this can be closed as done

PR #89, opened 2023-01-06
Enums (Scala 3)

keynmol:
Relates to #72 

I say "relates" not "closes" because there are restrictions in what can be part of enums, and I need to run it on a few more codebases to


eed3si9n (approved):
Thanks, @keynmol!

PR #90, opened 2023-01-07
Use macOS for testing instead

eed3si9n:
Problem
-------
For some reason code generation doesn't seem to work on Linux, thus PRs without the generated code fails the test.

Solution
--------
Switch to using macOS.


eed3si9n:
Self-merging this to unblock pending PRs.

PR #91, opened 2023-01-07
chore: bump dotty compiler expected

ckipp01:
This is way higher as of https://github.com/tree-sitter/tree-sitter-scala/pull/82
so we'll bump this in CI now.


ckipp01:
Superseded by https://github.com/tree-sitter/tree-sitter-scala/pull/93.

PR #92, opened 2023-01-07
fix: remove hidden arrow node

ckipp01:
This reverts a change made in
https://github.com/tree-sitter/tree-sitter-scala/commit/fa8d64b77ed9a53067317eb69725b61bf2c634b1#diff-919ac210accac9ecc55a76d10a7590e3d85ca3f0e165b52d30f08faee486d0cbL140-R142
where we introduced `_arrow`. This change will break the current neovim
highlights and we don't need it to be a named node.

refs: https://github.com/nvim-treesitter/nvim-treesitter/pull/4113

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -144,7 +144,7 @@ module.exports = grammar({
 
     renamed_identifier: $ => seq(
       field('name', $.identifier),
-      choice($._arrow, 'as'),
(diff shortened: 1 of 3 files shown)


eed3si9n (approved):
My bad.
@ckipp01 Thanks for the fix!

PR #93, opened 2023-01-07
Fix grep to detect C changes

eed3si9n:
Problem
-------
My grep has a bug, so *.sh file changes are detected as C change.

Solution
--------
1. Add backslash before dot.
2. Bump up the expected Dotty score.


ckipp01 (approved):
LGTM!

PR #94, opened 2023-01-07
chore(parser): update parser to include latest changes.

ckipp01:
The commit this will be generated from is ef3c8305d82a0a12bc4118ab2088353b49a10509.

PR #95, opened 2023-01-07
Fix the diff change detection 

eed3si9n:
Ref https://github.com/tree-sitter/tree-sitter-scala/issues/85
Ref https://github.com/tree-sitter/tree-sitter-scala/pull/94

Problem
-------
Currently the diff change detection is broken because it fails
when the changes are empty.

Solution
--------
1. Fix the logic.
2. Include the file names in the error message.
3. Run diff

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,14 +69,15 @@ jobs:
           npm run build
           npm test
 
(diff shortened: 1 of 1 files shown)


ckipp01 (approved):
Thanks for fixing this!

Issue #73, opened 2023-01-05
Given instances


eed3si9n:
https://docs.scala-lang.org/scala3/reference/contextual/givens.html

```scala
given intOrd: Ord[Int] with
  def compare(x: Int, y: Int) =
    if x < y then -1 else if x > y then +1 else 0

given listOrd[T](using ord: Ord[T]): Ord[List[T]] with

  def compare(xs: List[T], ys: List[T]): Int = (xs, ys) match
    case (Nil, Nil) => 0
    case (Nil, _) => -1
    case (_, Nil) => +1
    case (x :: xs1, y :: ys1) =>
      val fst = ord.compare(x, y)
      if fst != 0 then fst else compare(xs1, ys1)
```

PR #99, opened 2023-01-08
Given instances

eed3si9n:
Fixes https://github.com/tree-sitter/tree-sitter-scala/issues/73

This implements support for givens instances.


script/smoke_test.sh
```diff
@@ -4,7 +4,7 @@
 
 SCALA_SCALA_LIBRARY_EXPECTED=87
 SCALA_SCALA_COMPILER_EXPECTED=45
-DOTTY_COMPILER_EXPECTED=56
+DOTTY_COMPILER_EXPECTED=55
```
eed3si9n, line 7: Note 1% regression on Dotty compiler.

ckipp01: Odd, I'd expect this to bump the coverage up, not down. Any idea why?

eed3si9n: With these small percentage difference, it's hard to tell. Potentially some files were parsed incorrectly before, and now that givens is interpreted correctly it's hitting some other unsupported feature?

ckipp01 (approved):

PR #100, opened 2023-01-08
Bump tree sitter to 0.20.7 and expose highlights query in rust binding

VixieTSQ:
Opened in favor of #55. Updates tree-sitter-cli in package.json and tree-sitter inside rust binding. Also uncomments line that exposes the `highlights` query in the rust binding.


eed3si9n (commented):
Overall the change LGTM.
Could you exclude `src/grammar.json` out of the PR please?
Per #85, the code generation is no longer sync'ed with the PR, and it currently requires 34GB of RAM to successfully generate C code (at least on Linux).

VixieTSQ:
> Overall the change LGTM. Could you exclude `src/grammar.json` out of the PR please? Per #85, the code generation is no longer sync'ed with the PR, and it currently requires 34GB of RAM to successfully generate C code (at least on Linux).

done!

eed3si9n:
> done!

I meant excluding from the change by doing `git checkout src/`, as opposed to deleting it.

VixieTSQ:
> > done!
> 
> I meant excluding from the change by doing `git checkout src/`, as opposed to deleting it.

I think git hates me right now... Either way I tried to add back the old version of src/grammar.json
I might have messed up? I can't tell. This is weirdly painful

eed3si9n:
ok. I created https://github.com/tree-sitter/tree-sitter-scala/pull/101.

PR #101, opened 2023-01-08
Bump tree sitter to 0.20.7 and expose highlights query in rust binding, take 3

eed3si9n:
This is a resend of #100 minus a generated file.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "tree-sitter-scala",
-  "version": "0.19.1",
+  "version": "0.20.7",
   "description": "Scala grammar for tree-sitter",
   "main": "bindings/node",
   "keywords": [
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "tree-sitter-scala",
-  "version": "0.19.1",
+  "version": "0.20.7",
```
ckipp01, line 3: Aren't these versions independent of each other?  Meaning we bumped the one down below, but that shouldn't necessary bump this is also to 0.20.7 since this is our release version. Shouldn't this just be the semver version of tree-sitter-scala?

eed3si9n: Yea. The whole situation around versioning these grammar is murky to me. It's as-if grammars evolve at its own pace but the version is given at the timing of tree-sitter release? 
It would also be convenient to know which grammar is expected to work with which tree-sitter version if there's some dependency, so the version might be functioning for the marker as well (like `_2.12` suffix in Scala world).

Issue #97, opened 2023-01-07
Investigate memory usage of parser generator


keynmol:
[We flew too close to the sun](https://github.com/tree-sitter/tree-sitter/issues/1890#issuecomment-1374577925)

Generation takes ~34GB of RAM, and that's why it silently fails on CI - NPM just swallows the SIGKILL.

I confirmed it on my machine by running the Rust binary directly.

We should see what causes such explosion of states (or whatever) and try to either fix TS itself, or rein in the grammar's appetite.

eed3si9n:
My System76 desktop from 2014 (with 32GB RAM) used to be able to generate `src/parser.c` until yesterday, but post-opaque type or enum, it seems like it has lost the ability to do so as well. I am glad I Newegg'ed when you shared us the magical 34GB number.

keynmol:
Didn't you have the same issue on your Using clauses PR? Perhaps less memory, but already above the 8G GitHub runners can afford


The inline PR is likely to get much worse, so keep that Newegg tab open

eed3si9n:
```
$ git co f504a9a63b0310c16418b70a6d2d65f1df8e8c5b
....

HEAD is now at f504a9a Merge pull request #93 from eed3si9n/wip/bump_score
$ /usr/bin/time -f "%M" node_modules/.bin/tree-sitter generate
27371636
```

so earlier today the memory requirement was 27GB.

eed3si9n:
Here's a PR that brings down the memory usage to 11GB - https://github.com/tree-sitter/tree-sitter-scala/pull/102

PR #102, opened 2023-01-08
Optimize codegen memory usage

eed3si9n:
Fixes https://github.com/tree-sitter/tree-sitter-scala/issues/97

Problem
-------
Currently codegen uses 34GB, which is more than my desktop can handle. ahlinc kindly let us know about `--report-states-for-rule` flag to try to identify some choke points.
To my surprise `class_definition` uses 3728 states - https://gist.github.com/eed3si9n/009b1e5f11f5a80ea2c98d030bd25e13

Solution
--------
This adds intermediate node called `_class_constructor`, which is right associated to deal with the modifiers, but now `class_definition` can be left associated.

After this change:
```
$ /usr/bin/time -f "%M" node_modules/.bin/tree-sitter generate
11723152
```
This brings down memory usage to 11GB.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -219,25 +219,31 @@ module.exports = grammar({
       field('body', optional($.template_body)),
     )),
 
-    class_definition: $ => prec.right(seq(
+    class_definition: $ => prec.left(seq(
       repeat($.annotation),
       optional($.modifiers),
       optional('case'),
       'class',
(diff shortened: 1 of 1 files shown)


ckipp01 (approved):
Great job hunting this down @eed3si9n!

LGTM

PR #103, opened 2023-01-08
Sync and improve highlighting queries, add Scala 3 specific constructs and tests

### LINKED ISSUE #98: Add highlight queries for enums
Even though they are now parsed correctly, enum keywords are missing from highlight queries 
image

Please ignore the rest of poor highlighting - it's an artifact of my botched local setup.

keynmol:
closes #98 

The highlighting queries in this repo are out of date with the version in nvim-treesitter: https://github.com/nvim-treesitter/nvim-treesitter/blob/master/queries/scala/highlights.scm

That version has some issues, so I used it as the bases and moved/simplified/added rules.
In particular we are more precise about highlighting identifiers (before they would be aggressively highlighted as variable in lots of places)

### plans

I feel like we should grow the test coverage of highlighting queries, and once we're satisfied - upstream the changes

### using these queries

```
cp queries/* ~/.local/share/nvim/site/pack/packer/start/nvim-treesitter/queries/scala/
```

The difference are subtle (and some identifiers have the correct highlighting only because they're capitalised), but in particular class/enum parameters and trait/class/object/enum names are treated correctly.

nvim-treesitter version 
image

this version
image

### DIFF
--- a/test/highlight/basics.scala
+++ b/test/highlight/basics.scala
@@ -1,0 +1,37 @@
+object Hello {
+// ^ keyword 
+//       ^ type.definition
+  val x = if (true) (25 * 1.0) else "hello"  
+// ^keyword
+//    ^variable
+//         ^conditional
+//             ^boolean
+//                   ^number
+//                        ^float    
+//                                    ^string
+//                              ^conditional 
+
+  trait Test {
+// ^ keyword 
+//       ^ type.definition
+     def meth(i: Int)(implicit x: Boolean) = ???
+//    ^keyword.function  
+//                       ^keyword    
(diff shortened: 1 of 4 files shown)


test/highlight/basics.scala
```diff
@@ -0,0 +1,37 @@
+object Hello {
```
keynmol, line 1: Tree sitter has builtin support for unit testing highlight rules: https://tree-sitter.github.io/tree-sitter/syntax-highlighting#unit-testing

eed3si9n (approved):
Thanks, @keynmol!

PR #105, opened 2023-01-08
Make smoke test script better for diffing results + restore removed highlight

keynmol:
Inspired by #99 dropping Dotty coverage, I wanted to have a way to diff results of two branches.

This does the following:

1. Accept a `PRODUCE_REPORTS` env variable which instructs the script to dump detailed report into a file
2. Format the output in a way that makes it easier for diffing tools .

For example, I have a git worktree of master branch checked out, so I can run the script on both branches and compare them using

> git diff --no-index report-dotty-compiler.txt worktree-master/report-dotty-compiler.txt

or use something like delta.

---

Also restores a query highlight that shouldn't have been removed


.github/workflows/ci.yml
```diff
@@ -62,15 +62,15 @@ jobs:
         run: gcc test/test-stack.c -o a.out && ./a.out
 
       - name: Generate parser
-        if: ${{ runner.os == 'macOS' }}
+        if: ${{ runner.os != 'Windows' }}
```
eed3si9n, line 62: Do we know that the parser is freshly generated for Linux, and now using whatever that's already there?

script/smoke_test.sh
```diff
@@ -21,10 +21,18 @@ failed=0
...
-  echo "$out"
+  local name=$3
+  cmd="npm exec -c 'tree-sitter parse $source_dir/**/*.scala --quiet --stat' | sort | sed 's%$source_dir%%g'"
+  echo
+  echo "Parse $source_dir: $cmd"
+  out=$((eval $cmd) || true)
+
+  if [ ! -e "$PRODUCE_REPORTS" ]; then
+    echo "$out" | sed G | sed -E 's/([0-9]+) ms//' | grep -v 'success percentage' > "report-$name.txt" 
+  fi
```
eed3si9n, line 34: 👍🏼

.github/workflows/ci.yml
keynmol: I think with your recent PR bringing down memory usage we can assume it _is_ being generated and then revert if it's not the case?

ckipp01:
Not a huge deal at all, but now we do get double reporting for both mac and linux on the summary page. Not really a blocker, but is it necessary?

.github/workflows/ci.yml
ckipp01: Looking at the time difference now I'm assuming that generation on linux _is_ happening since before it was taking like < 20 second and now 2 minutes.

keynmol:
Oh right, good point.



> Not a huge deal at all, but now we do get double reporting for both mac and linux on the summary page. Not really a blocker, but is it necessary?

Good point, we can just go back to only running those stages on Linux as it's the fastest?

ckipp01:
> Oh right, good point.
> 
> > Not a huge deal at all, but now we do get double reporting for both mac and linux on the summary page. Not really a blocker, but is it necessary?
> 
> Good point, we can just go back to only running those stages on Linux as it's the fastest?

Yea sure that makes sense! Do you want to do that right away or should we merge as it and do it in another pr?

ckipp01 (approved):
Vamos!

PR #106, opened 2023-01-09
Optimize codegen memory usage, part 2

eed3si9n:
Ref #97

Problem
-------
Memory usage already regressed a bit to 15GB.

```
$ /usr/bin/time -f  "%P %M" node_modules/.bin/tree-sitter generate
99% 15625000
```

Running

```
$ node_modules/.bin/tree-sitter generate --report-states-for-rule compilation_unit
given_definition              	4434
type_definition               	2538
function_definition           	2322
```

show that `given_definition` might be the culprit.

Solution
--------
Apply the same `class_constructor` trick to `given_definition`, enum, `function_definition`, `function_declaration`, bringing down the memory usage down to one (1) GB.

```
$ /usr/bin/time -f "%P %M" node_modules/.bin/tree-sitter generate
99% 1078896
```

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -98,10 +98,7 @@ module.exports = grammar({
 
     enum_definition: $ => seq(
       'enum',
-      field('name', $.identifier),
-      field('type_parameters', optional($.type_parameters)),
-      optional($.access_modifier),
-      field('class_parameters', repeat($.class_parameters)),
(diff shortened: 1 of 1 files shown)


ckipp01 (approved):
Wow, great job @eed3si9n !

LGTM

PR #107, opened 2023-01-09
Improve expression hierarchy

eed3si9n:
This is on top of https://github.com/tree-sitter/tree-sitter-scala/pull/106

Problem
-------
Currently we use `$.expression` everywhere even though it often
requires much narrower set of expressions.

Solution
--------
Now that we've gained some memory budget, this adds
`_simple_expression`, mirroring the EBNF, and adjusts
`infix_expression` etc to use narrower choice of expression tokens.
With this change, the smoke test for Scala 2 library jumps to 95%.


script/smoke_test.sh
```diff
@@ -2,9 +2,9 @@
 
 # This is an integration test to generally check the quality of parsing.
 
-SCALA_SCALA_LIBRARY_EXPECTED=87
-SCALA_SCALA_COMPILER_EXPECTED=45
-DOTTY_COMPILER_EXPECTED=55
+SCALA_SCALA_LIBRARY_EXPECTED=95
+SCALA_SCALA_COMPILER_EXPECTED=66
+DOTTY_COMPILER_EXPECTED=56
```
eed3si9n, line 7: 95, 66, 56.

ckipp01: Awesome to see this keep going up.

ckipp01 (approved):
This looks good. LGTM!

PR #108, opened 2023-01-09
Fixes givens

eed3si9n:
This is on top of https://github.com/tree-sitter/tree-sitter-scala/pull/106

Problem
-------
Current grammar doesn't handle a common givens pattern:

```scala
given Context = ctx
```

Solution
--------
This fixes it by denoting name etc optional.


ckipp01 (approved):
Looks good, LGTM!

PR #112, opened 2023-01-09
Vararg class params and operator identifiers

keynmol:
This does bring up the coverage, but unfortunately this handling of operator_identifier caused issues in prefix_expression tests.

I'm not sure how to fix the tests.

Another option would be to perhaps do `_identifier_like: $ => choice($.identifier, $.operator_identifier)` but this takes us further away from what Scala syntax is like.

### DIFF
--- a/corpus/definitions.txt
+++ b/corpus/definitions.txt
@@ -858,3 +858,67 @@ class B {
           (class_parameter
             (identifier)
             (type_identifier)))))))
+
+
+
+===================================
+Operator identifiers
+===================================
+
+type ::[+Ab] = scala.collection.immutable.::[Ab]
+val :: = scala.collection.immutable.::
+
+val +: = scala.collection.+:
+val :+ = scala.collection.:+
+
+val test = id.##
+
+---
+
+(compilation_unit
+  (type_definition
+    (type_identifier)
+    (type_parameters
+      (covariant_type_parameter
+        (identifier)))
+    (infix_type
+      (stable_type_identifier
+        (stable_identifier
+          (identifier)
+          (identifier))
+        (type_identifier))
+      (operator_identifier)
+      (generic_type
+        (type_identifier
+          (MISSING _plainid))
+        (type_arguments
+          (type_identifier)))))
+  (val_definition
+    (operator_identifier)
+    (postfix_expression
+      (field_expression
+        (field_expression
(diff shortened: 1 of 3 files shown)


corpus/definitions.txt
```diff
@@ -858,3 +858,67 @@ class B {
...
+    (infix_type
+      (stable_type_identifier
+        (stable_identifier
+          (identifier)
+          (identifier))
+        (type_identifier))
+      (operator_identifier)
+      (generic_type
+        (type_identifier
+          (MISSING _plainid))
```
keynmol, line 893: I honestly don't know what this is.

Found https://github.com/tree-sitter/tree-sitter/issues/1043 but not sure I understand both the issue and a way to fix it..

grammar.js
```diff
@@ -1045,7 +1048,7 @@ module.exports = grammar({
 
     wildcard: $ => '_',
 
-    operator_identifier: $ => /[^\s\w\(\)\[\]\{\}'"`\.;,]+/,
+    operator_identifier: $ => /[!#%&*+-\/:<=>?@'^\|‘~]+/,
```
keynmol, line 1051: This is a more precise match, from https://docs.scala-lang.org/scala3/reference/syntax.html#

What it _doesn't_ match is unicode characters from Sm, So, that's why in Predef this is an error:

```scala
    @deprecated("Use `->` instead. If you still wish to display it as one character, consider using a font with programming ligatures such as Fira Code.", "2.13.0")
    def →[B](y: B): (A, B) = ->(y)
```

keynmol: I don't want to include all the characters from those character sets (thousands), so I think we can live with the limitation that ligatures will cause errors in the grammar

keynmol: People shouldn't be using ligatures anyways. ASCII or die.

script/smoke_test.sh
```diff
@@ -2,9 +2,9 @@
 
 # This is an integration test to generally check the quality of parsing.
 
-SCALA_SCALA_LIBRARY_EXPECTED=95
+SCALA_SCALA_LIBRARY_EXPECTED=98
```
eed3si9n, line 5: 🚀

eed3si9n (approved):
Thanks, @keynmol!

Issue #114, opened 2023-01-09
Self type


eed3si9n:

eed3si9n:
https://github.com/tree-sitter/tree-sitter-scala/commit/6990b2ae017784266e2bee43c9cf841e42c817af

PR #115, opened 2023-01-09
Scala 2/3: Self-types

eed3si9n:
Fixes https://github.com/tree-sitter/tree-sitter-scala/issues/114

Picking off a commit from @keynmol's branch.

### DIFF
--- a/corpus/definitions.txt
+++ b/corpus/definitions.txt
@@ -785,3 +785,45 @@ enum Hello[Y] extends java.Enumeration derives Codec, Eq {
       )
     )
   )
+
+================================
+Self types
+================================
+
+trait A {
+  self => 
+  def f: Int
+}
+
+class B {
+  self: Something[A] => 
+
+  case class Hello(a: Int)
+}
+
+---
+
+(compilation_unit
+  (trait_definition
+    (identifier)
+    (template_body
+      (self_type
+        (identifier))
+      (function_declaration
+        (identifier)
+        (type_identifier))))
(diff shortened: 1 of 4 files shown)


eed3si9n:
Smoke test caught that this regresses on parsing the following apprently:

```scala
private[typechecker] def isFullyDefined(tp: Type): Boolean =
  tp match {
    case TypeVar(_, constr) if constr.inst == NoType => false
  }
```

eed3si9n:
Actually I think that particular pattern seems to be broken on master as well? I think we should take the hit on Scala 2 compiler regression and merge this as-is.

ckipp01:
> Actually I think that particular pattern seems to be broken on master as well? I think we should take the hit on Scala 2 compiler regression and merge this as-is.

I'm fine with that.

Also regarding what @keynmol said in the other pr:

> For this and future changes may I suggest adding highlighting tests as well to https://github.com/tree-sitter/tree-sitter-scala/blob/master/test/highlight/scala3.scala?

I don't want to press, since you're doing all the work here 😆, but should we add these, or tackle them later? If you don't want to I'm happy to try and go back through and add some when. I get some more time.

eed3si9n:
> Also regarding what @keynmol said in the other pr:
> 
> > For this and future changes may I suggest adding highlighting tests as well to https://github.com/tree-sitter/tree-sitter-scala/blob/master/test/highlight/scala3.scala?

Sure. I added highlight for self-type.

PR #116, opened 2023-01-09
Include operator-like identifier as simple expression

eed3si9n:
Problem
-------
Currently operator-like identifiers are not part of the simple expression, so you can't call `::(123)`.

Solution
--------
This includes `$.operator_identifier` into `_simple_expression`, and further improves the expression hierarchy.

Note that this removes test on double-prefix-expression. It's not allowed in Scala spec to have double-prefix.


script/smoke_test.sh
```diff
@@ -4,7 +4,7 @@
 
 SCALA_SCALA_LIBRARY_EXPECTED=98
 SCALA_SCALA_COMPILER_EXPECTED=63
-DOTTY_COMPILER_EXPECTED=63
+DOTTY_COMPILER_EXPECTED=70
```
keynmol, line 7: 🚀 🚀 🚀 🚀

ckipp01: Noice!

ckipp01 (approved):

PR #117, opened 2023-01-09
Scala 3 enums: fix simple case inheritance

keynmol:
Seems like I missed it originally.

Now the reference enum looks _perfect_: 
image


eed3si9n (approved):
Thanks!

PR #118, opened 2023-01-10
Spec-correct identifier

eed3si9n:
This is on top of https://github.com/tree-sitter/tree-sitter-scala/pull/116

Problem
-------
The `identifier` token is hacked together, so we never quite get the
parsing right.

Solution
--------
My friend Ethan Atkins at some point experimented with generating
tree-sitter grammar out of EBNF (https://github.com/eatkins/tree-sitter-ebnf-generator/tree/master/examples/scala)
and his identifier, as far as I know is spec-correct.
We simplified the regex a bit to fit into the tree-sitter's unsigned
short int count, but it should be comparable.


grammar.js
```diff
@@ -1035,24 +1044,30 @@ module.exports = grammar({
...
+      $._backquoted_id,
+    )),
+
+    /**
+     * alphaid          ::=  upper idrest
+     *                       |  varid
+     *
+     * The following regex was copied from Ethan's https://github.com/eatkins/tree-sitter-ebnf-generator/blob/b25508d4ed3c6f9926acce70b2a1d0847d45363a/examples/scala/grammar.js#L40-L41 minus the operator part, which we handle using operator_identifier.
+     */
+    _alpha_identifier: $ => /[\p{Lu}\p{Lt}\p{Nl}\p{Lo}\p{Lm}\$][\p{Lu}\p{Lt}\p{Nl}\p{Lo}\p{Lm}\$\p{Ll}_\u00AA\u00BB\u02B0-\u02B8\u02C0-\u02C1\u02E0-\u02E4\u037A\u1D78\u1D9B-\u1DBF\u2071\u207F\u2090-\u209C\u2C7C-\u2C7D\uA69C-\uA69D\uA770\uA7F8-\uA7F9\uAB5C-\uAB5F0-9]*(_[!#%&*+-\/:<=>?@'^\|‘~]+)?|[\p{Ll}_\u00AA\u00BB\u02B0-\u02B8\u02C0-\u02C1\u02E0-\u02E4\u037A\u1D78\u1D9B-\u1DBF\u2071\u207F\u2090-\u209C\u2C7C-\u2C7D\uA69C-\uA69D\uA770\uA7F8-\uA7F9\uAB5C-\uAB5F_][\p{Lu}\p{Lt}\p{Nl}\p{Lo}\p{Lm}\$\p{Ll}_\u00AA\u00BB\u02B0-\u02B8\u02C0-\u02C1\u02E0-\u02E4\u037A\u1D78\u1D9B-\u1DBF\u2071\u207F\u2090-\u209C\u2C7C-\u2C7D\uA69C-\uA69D\uA770\uA7F8-\uA7F9\uAB5C-\uAB5F0-9]*(_[!#%&*+-\/:<=>?@'^\|‘~]+)?/,
```
keynmol, line 1062: > (_[!#%&*+-\/:<=>?@'^\|‘~]+)

I can see the regex still contains the operator part - is the comment accurate then?

keynmol (approved):
Very nice - and covers all the cases where `_identifier` is supposed to be used!

grammar.js
eed3si9n: Yea. Note that symbolic characters follow mandatory underscore, e.g. for `unary_!`.

PR #120, opened 2023-01-10
Extra highlights for scala 3 constructs

keynmol:


keynmol:
btw @eed3si9n  I think `given Typeable[T] with ...` doesn't parse - not sure why 
image

eed3si9n (approved):
Thanks!

PR #125, opened 2023-01-10
Export clauses

### LINKED ISSUE #76: Export clauses
https://docs.scala-lang.org/scala3/reference/other-new-features/export.html

```scala
class Copier:
  private val printUnit = new Printer { type PrinterType = InkJet }
  private val scanUnit = new Scanner

  export scanUnit.scan
  export printUnit.{status as _, *}
```

keynmol:
Closes #76 

### breaking highlighting

For consistency, I renamed import_selectors to namespace_selectors to share them with exports.
This will break with old queries.

That said, given how sensitive they are (any node/keyword not present in the grammar causes highlighting to break), we've broken them a while back.

We'll sync them to nvim-treesitter soon.

### DIFF
--- a/corpus/definitions.txt
+++ b/corpus/definitions.txt
@@ -185,7 +185,7 @@ import reflect.io.{Directory, File, Path}
     (identifier) (identifier))
   (import_declaration
     (identifier) (identifier)
-    (import_selectors (identifier) (identifier) (identifier))))
+    (namespace_selectors (identifier) (identifier) (identifier))))
 
 ================================
 Imports: Wildcard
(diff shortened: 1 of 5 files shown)


eed3si9n:
> That said, given how sensitive they are (any node/keyword not present in the grammar causes highlighting to break), we've broken them a while back.

While implementing Scala 3 macros symbol literal got subsumed by quote expression, but I added back a dummy node called symbol_literal so it doesn't break the existing query. Adding unconnected node hopefully isn't too expensive.

eed3si9n (approved):
Thanks!

Issue #127, opened 2023-01-10
SIP-44 - Fewer Braces support


eed3si9n:
https://docs.scala-lang.org/sips/fewer-braces.html

```scala
times(10):
  println("ah")
  println("ha")
```

```scala
credentials `++`:
  val file = Path.userHome / ".credentials"
```

```scala
xs.map:
  x =>
    val y = x - 1
    y * y
```

```scala
xs.map: x =>
  val y = x - 1
  y * y
```

## Compatiblity

**Note**: The existence of a newline changes the meaning in the following example:

```scala
$ scala
Welcome to Scala 3.3.0-RC1-bin-20230109-f56089b-NIGHTLY-git-f56089b (1.8.0_345, Java OpenJDK 64-Bit Server VM).
Type in expressions for evaluation. Or try :help.

scala> import language.experimental.fewerBraces

scala> def y(x: Any) = 1
def y(x: Any): Int

scala> val x = y:
     |   Int
     |
val x: Int = 1

scala> val x = y: Int
-- [E007] Type Mismatch Error: -------------------------------------------------
1 |val x = y: Int
  |        ^
  |        Found:    Any => Int
  |        Required: Int
  |
  | longer explanation available when compiling with `-explain`
1 error found
```

PR #128, opened 2023-01-10
SIP-44 Fewer braces support

eed3si9n:
Fixes https://github.com/tree-sitter/tree-sitter-scala/issues/127

This is on top of https://github.com/tree-sitter/tree-sitter-scala/pull/131

sip-44

Problem
-------
Currently our grammar does not support the fewer braces syntax, which basically lets us pass the last argument as a block after `:`.

Solution
--------
This implements fewer braces support for call_expression, infix_expression, with and without the lambda start.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -10,6 +10,7 @@ const PREC = {
   unit: 4,
   ascription: 4,
   postfix: 5,
+  colon_call: 5,
   infix: 6,
   constructor_app: 7,
   prefix: 7,
@@ -65,6 +66,7 @@ module.exports = grammar({
     [$.tuple_type, $.parameter_types],
     [$.binding, $._simple_expression],
     [$.binding, $.ascription_expression],
+    [$.binding, $._type_identifier],
     [$.if_expression, $.expression],
     [$.while_expression, $._simple_expression],
     [$.for_expression, $.infix_expression],
@@ -673,6 +675,7 @@ module.exports = grammar({
       $.tuple_type,
       $.stable_type_identifier,
       $._type_identifier,
+      $.wildcard,
     ),
 
     compound_type: $ => prec(PREC.compound, seq(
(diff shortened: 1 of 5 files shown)


keynmol:
This is going to sound very annoying, so brace yourself.

I took a look at the current state in the compilers:

3.2.2 (releases hours ago) still doesn't have it: 
image

3.3 will likely have it moved from experimental to general scope, but technically there are no guarantees and no stability promises, is there?

Do we want to make an unreleased syntax such a permanent part of the grammar at this point?

We seem to be only getting a modest bump on dotty codebase (I expected more tbh, need to dig more into the failures), and other than lots of experimental code I don't think there are benefits.

What do you think?

eed3si9n:
> Do we want to make an unreleased syntax such a permanent part of the grammar at this point?

The syntax was approved August 2022 (https://contributors.scala-lang.org/t/results-of-the-first-sip-meeting-of-the-scala-3-era/5898) after seemingly endless round of debates on whether we want it or not. As I see it the built-in delay between it being approved and actual release cycle is partly so tooling (including tree-sitter) can use it to implement support for it, so if people rely on tooling working it won't go belly up every time a new feature is added, which many people have complained, and it's getting better.

keynmol:
> so tooling (including tree-sitter) can use it to implement support for it,

Ok, I think this is key, and I agree with it. As an additional datapoint - Scalameta supports it since september: https://github.com/scalameta/scalameta/pull/2848

grammar.js
```diff
@@ -954,9 +957,38 @@ module.exports = grammar({
...
+          $.wildcard,
+        ),
+        '=>',
+      ))),
+      $._colon_argument_block,
+    )),
+
+    _colon_argument_block: $ => prec.left(seq(
+      $._indent,
+      field('body', $._block),
```
keynmol, line 990: Should we have `case_block` alternative here?

Seems to be allowed:

```scala
scala> import language.experimental.fewerBraces

scala> List(1,2,3,4,5).collect:
     |   case i => i + 1
     |
val res0: List[Int] = List(2, 3, 4, 5, 6)
```

eed3si9n: Nice catch!
```
  ColonArgument     ::=  colon [LambdaStart]
                         (CaseClauses | Block)
```

eed3si9n: Done - https://github.com/tree-sitter/tree-sitter-scala/pull/128/commits/a7e939493522d04b4efad2ca57d0977c1513d330

keynmol (approved):
Excited about how non-invasive this turned out to be!
Amazing work.

PR #129, opened 2023-01-11
Setup nix with direnv

ghostbuster91:
This PR setups up a nix shell for the project. I think that when native libraries are involved it is especially beneficial to use nix to guarantee reproducibility with ease.

In addition to that it also configures `.envrc` file that can be read by `direnv` when entering the project which will automatically put the user into the nix shell.

Otherwise it has to be invoked manually by calling `nix-shell .`

Todo:
- [x] update readme


eed3si9n:
The description contains a todo. Is this ready for a review?

ghostbuster91:
@eed3si9n yes, I didn't want to spend time updating readme without knowing if you are ok in general with such approach. Sorry, I should've made it more clear. Let me know what you think and then I can update it.

eed3si9n:
I am not against adding a Nix setup, but is the general intent to replace the existing GitHub Actions steps or our local development process with Nix? I am kind of curious since I know of Nix, but I have never used it, and I am not sure what the tradeoffs are.

keynmol:
IMO - we don't need Nix on GHA because we want to rely on Tree Sitter's official bootstrap scripts which are all in the NPM package.

That said, it seems to be the thing-of-the-day to have a Nix setup for development, so I'm happy to just have it available and documented

ckipp01:
My two cents are basically the same. I don't mind having it here, with the hopes that a nix user will also maintain it. I wouldn't want to move CI to it or push that all users should have it locally to develop on this repo.

ghostbuster91:
My intention was only to make it slightly more convenient for nix user, not to enforce it on the rest. 

I have to admit that I haven't thought yet about using it on CI but @keynmol already answered that part :) If I might add two-cents here, it is all about making sure that the same version are used. For the development of grammar users needs two things node and c compiler. Nix takes care of that, but I agree that it is not that hard to ensure that manually.

Seems like we are on the same page, hence I will update the readme with the relevant section for nix-based users.

eed3si9n (approved):
Thanks, @ghostbuster91!

README.md
```diff
@@ -1,19 +1,24 @@
-tree-sitter-scala
-=================
+# tree-sitter-scala
```
ghostbuster91, line 1: My neovim did this change totally on his own, no idea why. Maybe it is not a recommended md syntax anymore?
I can revert it if you want.

PR #130, opened 2023-01-11
Add support for structural types

ghostbuster91:
I hope that it isn't too late to join the tree-sitter party :)

This PR adds supports for structural types which allows for using type projections on anonymous structural types that are usually used when working with a higher-kinded types without kind projector.

### DIFF
--- a/corpus/types.txt
+++ b/corpus/types.txt
@@ -320,3 +320,61 @@ opaque type B <: Test >: Help = String
     (lower_bound (type_identifier))
     (type_identifier))
 )
+
+==================================
+Structural type
+==================================
+
+type A = { def fly(): Unit }
+
+---
+
+(compilation_unit
+  (type_definition (type_identifier)
+    (structural_type
+      (function_declaration 
+        (identifier)
+        (parameters)
+        (type_identifier)
+      )
+    )
+  )
+)
+
+=========================================
+Anonymous structural type with projection
+=========================================
+
+type A = B[({ type f[x] = M[S, x] })#f]
+
+---
+
+(compilation_unit
+  (type_definition (type_identifier)
+    (generic_type
+      (type_identifier)
+      (type_arguments
+        (projected_type
+          (tuple_type
+            (structural_type
+              (type_definition
+                (type_identifier)
+                (type_parameters
+                  (identifier)
+                )
+                (generic_type
+                  (type_identifier)
+                    (type_arguments
+                      (type_identifier)
+                      (type_identifier)
+                    )
+                )
+              )
+            )
+          )
+          (type_identifier)
+        )
+      )
+    )
+  )
+)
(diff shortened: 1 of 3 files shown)


grammar.js
```diff
@@ -698,6 +699,8 @@ module.exports = grammar({
       field('return_type', $._type)
     )),
 
+    _structural_type: $ => $.template_body,
```
ghostbuster91, line 702: Not sure if it shouldn't be just inlined there.

eed3si9n: For later purpose of highlighting or folding, I wonder if this should be a visible node, that is `structural_type` without the underscore.

corpus/types.txt
```diff
@@ -320,3 +320,61 @@ opaque type B <: Test >: Help = String
...
+      )
+    )
+  )
+)
+
+=========================================
+Anonymous structural type with projection
+=========================================
+
+type A = B[({ type f[x] = M[S, x] })#f]
```
keynmol, line 348: _isn't Scala beautiful_

grammar.js
keynmol: For folding - sure, and this can just be aliases `_structural_type: $=>alias($.template_body, $.structural_type)` if this is the correct syntax

keynmol:
@ghostbuster91 looks good!

Could you please also add a couple of highlighting tests into https://github.com/tree-sitter/tree-sitter-scala/blob/master/test/highlight/basics.scala - nothing crazy, you can just use your test cases and ensure that keywords like `def` and types like `Unit` have the correct highlight groups

grammar.js
ghostbuster91: Right, there is also another way. We could make it an alias like `alias($.template_body, 'structural_type')`. 

So instead of having it a visible node which would render as:
```
(compilation_unit
  (type_definition (type_identifier)
    (structural_type
      (template_body
        (function_declaration 
          (identifier)
          (parameters)
          (type_identifier)
        )
      )
    )
  )
)
```
we would have:
```
(compilation_unit
  (type_definition (type_identifier)
      (structural_type
        (function_declaration 
          (identifier)
          (parameters)
          (type_identifier)
        )
    )
  )
)
```
wdyt?

ghostbuster91:
@keynmol sure, good idea :+1:

grammar.js
keynmol: I wonder if actually having the template_body there is better, because we can do generic highlights (like, "all function declarations in template_body should be marked with `method`, not `function`)

That said, I don't see any such queries there yet, and we can change them later.

So for now - happy to go with 

```
      (structural_type
        (function_declaration 
          (identifier)
          (parameters)
          (type_identifier)
        )
    )
```

because it reduces nesting in the tree.

eed3si9n (approved):
Thanks, @ghostbuster91!

PR #131, opened 2023-01-12
Fixes instance_expression

eed3si9n:
Problem
-------
Current grammar accepts expression immeidately after `new`, which is not correct, and doesn't work for Scala 3 syntax.

Solution
--------
This fixes it by actually using `$.template_body`. It does create a tricky conflict between `new A: <block>` construct and `new A: <type>`.
To deal with that, we are using `prec.dynamic`.


corpus/expressions.txt
```diff
@@ -428,8 +428,15 @@ Instance expressions
 
 class C {
   def main() {
-    a = new B
-    c = new D(e, f)
+    val a = new B
+    val c = new D(e, f)
+    val e = new E:
```
keynmol, line 433: For posterity, can you also add an example of the quiet syntax `new:`

```scala
scala> trait Hello:
     |   def x: Int
     |
// defined trait Hello

scala> val x: Hello = new:
     |   def x = 25
     |
val x: Hello = anon$1@6bd92538
```

eed3si9n: Done - https://github.com/tree-sitter/tree-sitter-scala/pull/131/commits/14a22e7baa5f26c810d0ff5865fa2fa83f24b0c3

keynmol (approved):
Nice

PR #138, opened 2023-01-12
Use storageclass

eed3si9n:
Problem
-------
Currently storage related modifiers use keyword.

Solution
--------
We should use `storageclass` for `private`, `protected`, and `inline`. See also https://www.sublimetext.com/docs/scope_naming.html#storage

> Keywords that affect the storage of a variable, function or data structure should use the following scope. Examples include static, inline, const, public and private.

### DIFF
--- a/queries/highlights.scm
+++ b/queries/highlights.scm
@@ -136,7 +136,6 @@
 ;; keywords
 
 (opaque_modifier) @type.qualifier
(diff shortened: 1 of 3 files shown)


keynmol (approved):
Verified that it works 
image

Personally I prefered keywords for inline modifiers, but I understand it's the right thing to do

Issue #56, opened 2022-07-07
`with` after extending abstract class with Param


domaspoliakas:
Hello 👋 

First time issue creator here, let me know if you need anything else from me!

The following code produces an error node:

```scala
abstract class Potato()
trait Tomato
class MyClass extends Potato() with Tomato
```

The treesitter AST shows the following:
```
class_definition [29, 0] - [29, 23]
  modifiers [29, 0] - [29, 8]
  name: identifier [29, 15] - [29, 21]
  class_parameters: class_parameters [29, 21] - [29, 23]
trait_definition [30, 0] - [30, 12]
  name: identifier [30, 6] - [30, 12]
class_definition [31, 0] - [31, 30]
  name: identifier [31, 6] - [31, 13]
  extend: extends_clause [31, 14] - [31, 30]
    type: type_identifier [31, 22] - [31, 28]
    arguments [31, 28] - [31, 30]
ERROR [31, 31] - [33, 0]
  identifier [31, 31] - [31, 35]
  identifier [31, 36] - [31, 42]
```

It seems that using `with` after extending an abstract class with parameters does not work.

EDIT: it seems that `abstract` class is not required, the same occurs with just a regular class too

eed3si9n:
Here's my PR for this - https://github.com/tree-sitter/tree-sitter-scala/pull/139

Issue #53, opened 2022-06-16
Allow compound types with constructors


ptwales:
```scala
trait A
abstract class B(b: String)
class C(c: String) extends B(c) with A
```

Output from nvim-treesitter/playground
```
trait_definition [0, 0] - [0, 7]
  name: identifier [0, 6] - [0, 7]
class_definition [1, 0] - [1, 27]
  modifiers [1, 0] - [1, 8]
  name: identifier [1, 15] - [1, 16]
  class_parameters: class_parameters [1, 16] - [1, 27]
    class_parameter [1, 17] - [1, 26]
      name: identifier [1, 17] - [1, 18]
      type: type_identifier [1, 20] - [1, 26]
class_definition [2, 0] - [2, 38]
  ERROR [2, 6] - [2, 36]
    identifier [2, 6] - [2, 7]
    class_parameters [2, 7] - [2, 18]
      class_parameter [2, 8] - [2, 17]
        name: identifier [2, 8] - [2, 9]
        type: type_identifier [2, 11] - [2, 17]
    extends_clause [2, 19] - [2, 31]
      type: type_identifier [2, 27] - [2, 28]
      arguments [2, 28] - [2, 31]
        identifier [2, 29] - [2, 30]
    identifier [2, 32] - [2, 36]
  name: identifier [2, 37] - [2, 38]
```

Referencing the second and third definitions from the subclass section in `definitions.txt`, the `extends_clause` should be a `compound_type` like:
```lisp
(class_definition
  (identifier)
  (class_parameters
     (class_parameter (identifier) (type_identifier)))
  (extends_clause
    (compound_type
      (type_identifier)
      (arguments (infix_expression (identifier) (operator_identifier) (identifier))
      (type_identifier))))
```
But I think `compound_type` should only take two arguments not three and `B(c)` should be one item not two but I don't know what it should be.

Also `with A` is marked as an identifier if it's on the same line as `extend` or an operator if it's on a newline.  I suspect that's not useful though but just in case.

eed3si9n:
Here's my PR for this - https://github.com/tree-sitter/tree-sitter-scala/pull/139

PR #139, opened 2023-01-13
Fixes extends clause

eed3si9n:
Fixes https://github.com/tree-sitter/tree-sitter-scala/issues/53
Fixes https://github.com/tree-sitter/tree-sitter-scala/issues/56

Problem
-------
1. In Scala 3, extends clause can be comma separated.
2. In Scala 2 or 3 extends clause can ctor with parameters:
   `class D with E(x) with F(y)`.

Solution
--------
`extends` is now followed by *constructor applications* separated by
either `with` or `,` as opposed to treating them as a compound type.

### DIFF
--- a/corpus/definitions.txt
+++ b/corpus/definitions.txt
@@ -333,6 +333,8 @@ object C {
         with C[T]
 }
 
+class D(c: String) extends E(c) with F
+
 ---
 
 (compilation_unit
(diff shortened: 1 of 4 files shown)


corpus/definitions.txt
```diff
@@ -333,6 +333,8 @@ object C {
         with C[T]
 }
 
+class D(c: String) extends E(c) with F
```
keynmol, line 336: Could you please add the original test case from #56:

```scala
class MyClass extends Potato() with Tomato
```

Which I believe tripped things over by having empty parens in there

keynmol (approved):
Nice!

corpus/definitions.txt
eed3si9n: Done - https://github.com/tree-sitter/tree-sitter-scala/pull/139/commits/ed65e637e1ced2def1de308faed3a2f82dcdf452

Issue #133, opened 2023-01-12
Full line of comments parsed as an `operator_identifier`


ckipp01:
Maybe an odd use case but I have a code base with comments like:

```
//////////////////
// MAIN MODULES //
//////////////////
```

And this ends up parsing as:

```
operator_identifier [24, 6] - [24, 27]
comment [25, 6] - [25, 27]
operator_identifier [26, 6] - [26, 27]
```

When it should really just be 3 comments.

eed3si9n:
It seems like it requires some code around it like:

```scala
val = x

/////////////////////
// Foo             //
/////////////////////
```

then `/////////////////////` gets highlighted as an operator.

eed3si9n:
Here's my PR for this - https://github.com/tree-sitter/tree-sitter-scala/pull/140

PR #140, opened 2023-01-13
Fix operator_identifier

eed3si9n:
Fixes https://github.com/tree-sitter/tree-sitter-scala/issues/133

Problem
-------
1. Currently `//` ends up matching `operator_identier`
2. There's also a bug in regex.

Solution
--------
1. Implement a workaround to avoid `//`.
2. Escape `-`, copying from the identifier regex.


script/smoke_test.sh
```diff
@@ -4,7 +4,7 @@
 
 SCALA_SCALA_LIBRARY_EXPECTED=100
 SCALA_SCALA_COMPILER_EXPECTED=66
-DOTTY_COMPILER_EXPECTED=74
+DOTTY_COMPILER_EXPECTED=66
```
ckipp01, line 7: Sort of a bummer, but it makes sense.

ckipp01 (approved):
LGTM

Issue #3, opened 2018-02-24
Singleton Types


mads-hartmann:
Currently a singleton type results in a

```
==================================
Singleton Types
==================================

class A { 
  def foobar: this.type = this
}

---

(compilation_unit 
  (class_definition 
    (identifier) 
    (template_body 
        (function_definition 
            (identifier) 
            (stable_type_identifier (identifier) (type_identifier)) 
            (identifier)))))

```

I'd prefer

```
(compilation_unit 
  (class_definition 
    (identifier) 
    (template_body 
        (function_definition 
            (identifier) 
            (stable_type_identifier (identifier) (singleton_type)) 
            (identifier)))))
```

eed3si9n:
Here's my PR for this - https://github.com/tree-sitter/tree-sitter-scala/pull/146

PR #146, opened 2023-01-14
Singleton type

eed3si9n:
Fixes https://github.com/tree-sitter/tree-sitter-scala/issues/3

Problem
-------
Currently singleton types are treated as `stable_type_identifier`.

Solution
--------
This adds `singleton_type`.


ckipp01 (approved):
Nice Job!

PR #147, opened 2023-01-15
feat: add in a job to automatically generate and sync daily

ckipp01:
I was about to sync and send in a pr again, but I figured why not just automate this. This change adds a new CI job that will run every morning and detect if there are changes. If so, it will commit them and send in a pr.

### DIFF
--- a/.github/workflows/sync.yml
+++ b/.github/workflows/sync.yml
@@ -1,0 +1,54 @@
+name: "Check changes and sync"
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: 0 5 * * *
+
+jobs:
+  check-and-sync:
+    runs-on: ubuntu-20.04
+    outputs:
+      all: ${{ steps.changes.outputs.all}}
+      c: ${{ steps.changes.outputs.c }}
+    steps:
+      - name: checkout tree-sitter-scala
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 10
+
+      - name: Generate parser from scratch
+        run: |
+          npm install
+          npm run build
+
+      - name: Check for changes
+        uses: tj-actions/verify-changed-files@v13
+        id: verify-changed-files
+        with:
+          files: |
+            src/grammar.json
+            src/node-types.json
+            src/parser.c
+            src/tree_sitter/parser.h
+
+      - name: Commit changes if necessary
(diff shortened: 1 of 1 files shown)


.github/workflows/sync.yml
```diff
@@ -0,0 +1,54 @@
...
+          files: |
+            src/grammar.json
+            src/node-types.json
+            src/parser.c
+            src/tree_sitter/parser.h
+
+      - name: Commit changes if necessary
+        if: steps.check-for-changes.outputs.files_changed == 'true'
+        run: |
+          git config user.name "GitHub"
```
keynmol, line 37: So github just provides an account you can use to do things like these?
Wild if so - I'm tempted to just merge this and run it manually to see if it works!

.github/workflows/sync.yml
```diff
@@ -0,0 +1,54 @@
...
+          LAST_COMMIT=$(git log --format-"%H" -n 1)
+          git commit -m "chore: generate and sync from $UPDATED_PARSERS"
+          git clean -xf
+
+      - name: Create Pull Request
+        if: steps.check-for-changes.outputs.files_changed == 'true'
+        uses: peter-evans/create-pull-request@v4
+        with:
+          title: "chore: generate and sync latest changes"
+          branch: generation
```
keynmol, line 49: where are you `git checkout -b` this branch? or does this GHA do it for you?

.github/workflows/sync.yml
ckipp01: Seems so, I took a lot of this from the way nvim-treesitter does this for their lock file.

.github/workflows/sync.yml
ckipp01: So I think technically we just commit it to master, but then the next step in `create-pull-request` specifies the `branch`. I think it must just take those changes, create a new branch, and then create the pr. Sort of copying this part from https://github.com/nvim-treesitter/nvim-treesitter/blob/5e9c6b6ba079a90a2545e651e3740b655b9ff035/.github/workflows/update-lockfile.yml#L40-L56.

eed3si9n (approved):
Thanks @ckipp01!

ckipp01:
Alright, I'm gonna merge and we'll try it out. I fully expect _something_ to not work right, but we'll see 😄

PR #150, opened 2023-01-15
chore: create an issue template

ckipp01:


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,63 @@
...
+                    pattern: (wildcard [1, 9] - [1, 10]))
+                  (lambda_expression [1, 15] - [2, 0]
+                    (identifier [1, 15] - [1, 16])
+                    (integer_literal [1, 20] - [1, 21])))))))
+        ```
+    validations:
+      required: true
+
+  - type: textarea
+    id: expectation
```
ckipp01, line 54: I made the code sample and the error required, but I figured I'd keep this one optional. Some people may have no idea what it should be, so I don't think we need to force this here. Just getting the actual error sample and code is good.

keynmol: Often I myself have no idea what I expect, it's usually "something but not _that_"

ckipp01: Me and you both.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,63 @@
...
+          }
+        ```
+    validations:
+      required: true
+
+  - type: textarea
+    id: error
+    attributes:
+      label: Show the error node
+      description: Show us the output and where the ERROR node is appearing
```
keynmol, line 33: I think we need to put somewhere instructions on how to obtain this output

ckipp01: Yea I was thinking about putting a section in the contributing docs about obtaining it. Let me create a section in there and add it to https://github.com/tree-sitter/tree-sitter-scala/pull/149 and then I'll link it from here.

ckipp01: Alright, did that and updated it.

eed3si9n (approved):

PR #151, opened 2023-01-15
chore: remove unused examples

ckipp01:
These haven't been touched in 5 years and they don't seem to be used at
all. Let's just ditch them.


keynmol:
Actually, we shouldn't be removing them - we need to add Scala 3 code to them :D https://github.com/tree-sitter/tree-sitter-scala/blob/master/package.json#L20
They're part of our CI

ckipp01:
pff sorry, should have noticed that 😆

PR #152, opened 2023-01-15
fix: ensure check for changed file uses id, not name

ckipp01:


ckipp01:
I'm just going to quickly merge this to try and get the automation to work. It was just a silly mistake on my end using the name of the job instead of the id.

Issue #159, opened 2023-01-16
Crash


eed3si9n:
@ckipp01 @eed3si9n I went back in time trying different commits. The bug seemed to have been introduced in this commit that has some suspicious code: https://github.com/tree-sitter/tree-sitter-scala/commit/1e70f7de9ee42247e8922ab1de3960582e78f479

_Originally posted by @gvolpe in https://github.com/nvim-treesitter/nvim-treesitter/issues/4170#issuecomment-1384526588_

### Commit of tree-sitter-scala you tested this on

fce8f8c3f6210a3eaf7ca210a1c0ebacc974ddda

### A code sample showing the error

```scala
object Main
```

### Show the error node

```
#0  0x000055e0c16039a0 in ts_decode_utf8 ()
#1  0x000055e0c160302e in ts_lexer.get_column ()
#2  0x00007f4a9064dfbe in tree_sitter_scala_external_scanner_scan () from /home/stephan/.local/share/nvim/site/pack/packer/start/nvim-treesitter/parser/scala.so
#3  0x000055e0c15e8afd in ts_parser_parse ()
#4  0x000055e0c13b7288 in parser_parse (L=0x7f4a937b1380) at /home/stephan/projects/neovim/src/nvim/lua/treesitter.c:421
#5  0x000055e0c163c1b6 in lj_BC_FUNCC ()
#6  0x000055e0c16261da in lua_pcall (L=0x20, nargs=<optimized out>, nresults=<optimized out>, errfunc=<optimized out>) at lj_api.c:1145
#7  0x000055e0c13af1b2 in nlua_pcall (lstate=0x7f4a937b1380, nargs=2, nresults=1) at /home/stephan/projects/neovim/src/nvim/lua/executor.c:153
#8  0x000055e0c13afc97 in nlua_call_ref (ref=5, name=0x55e0c16b8278 "buf", args=..., retval=true, err=0x7ffff60bd558) at /home/stephan/projects/neovim/src/nvim/lua/executor.c:1559
#9  0x000055e0c128d397 in decor_provider_invoke (ns_id=1, name=0x55e0c16b8278 "buf", ref=5, args=..., default_true=true, perr=0x55e0c17dcb30 <provider_err>) at /home/stephan/projects/neovim/src/nvim/decoration_provider.c:36
[...]

PR #160, opened 2023-01-16
Fixes buffer overrun

eed3si9n:
Fixes https://github.com/tree-sitter/tree-sitter-scala/issues/159
Ref https://github.com/nvim-treesitter/nvim-treesitter/issues/4170

Problem
-------
The buffer given for payload serialization is 1024, but we are currently using (1024 + 3) * sizeof(int)

Solution
--------
Reduce it way down.

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -1,0 +1,25 @@
+name: Fuzz parser
+
+# Run this workflow on changes to the external scanner
+on:
+  workflow_dispatch:
+  #push:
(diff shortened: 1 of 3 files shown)


ckipp01 (approved):
👍🏼 let's hope this gets rid of the pesky crashes.

ckipp01:
@keynmol I'm going to just merge and try to update nvim-treesitter right away with this after it generates since people are hitting on this. Please do still take a gander at this though and we can adjust if necessary.

PR #164, opened 2023-01-17
Fixes crash, take 2

eed3si9n:
Fixes https://github.com/tree-sitter/tree-sitter-scala/issues/159

Problem
-------
tree-sitter-scala crashes under mysterious conditions. gdb seems to indicate that it happens around `get_column`.

Solution
--------
Avoid calling `get_column` when it's at the end of file. That seems to satisfy fuzz.

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -3,14 +3,14 @@ name: Fuzz parser
 # Run this workflow on changes to the external scanner
 on:
   workflow_dispatch:
-  #push:
-  #  paths:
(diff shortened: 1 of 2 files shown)


theHamsta:
This get_column at eof hit caused many of the segfaults we saw before in parsers

ckipp01 (approved):
Nice find! Great to see the fuzzer working as well.

LGTM!

gvolpe:
Can confirm this actually fixes the issue, thanks again @eed3si9n ! 😍

Issue #136, opened 2023-01-12
Test against nvim-treesitter highlights


keynmol:
To ensure we don't break highlights in nvim-treesitter repo, I suggest we do the following:

1. On PR builds, clone nvim-treesitter
2. Replace local queries with versions from nvim-treesitter
3. Run tests 

This should help us catch regressions and act accordingly.

We only need to do it because parser/highlight syncing is not atomic, so they can go out of sync.

keynmol:
Done in #169

PR #169, opened 2023-01-18
feat(ci): test against nvim-treesitter highlights

ckipp01:
This PR starts testing against the highlights that nvim-treesitter has
defined to ensure we don't break things without warning. The general
flow is as follows:

- When on linux we clone nvim-treesitter
- After all of our initial testing we copy the queries that they have
  defined
- If they differ, we run our highlight tests against them
- We also issue a warning that the highlights have diverged (more than
  likely this will fail the highlighting tests anyways)

This is in ref to https://github.com/tree-sitter/tree-sitter-scala/issues/136. @keynmol let me know if this is what you had in mind.


.github/workflows/ci.yml
```diff
@@ -36,7 +36,7 @@ jobs:
           fetch-depth: 10
 
       - name: checkout scala/scala
-        if: ${{ runner.os != 'Windows' }}
+        if: ${{ runner.os == 'Linux' }}
```
ckipp01, line 39: Not exactly related to this but I realized we are cloning this on mac as well, but not running the smoke tests on mac. So I just switched this back to only clone on the os that we actually run it on.

.github/workflows/ci.yml
```diff
@@ -90,3 +97,23 @@ jobs:
...
+        with:
+          files: |
+            queries/highlights.scm
+
+      - name: Test quries if out of sync with nvim-treesitter
+        if: steps.verify-changed-files.outputs.files_changed == 'true'
+        run: |
+            echo "::warning file=queries/highlights.scm::Queries in this repo are out of sync with nvim-treesitter"
+            git diff queries/highlights.scm
+            npm run test
```
keynmol, line 119: Pretty sure the answer is "yes", but for discussion sake: _should_ we necessarily fail the job if the queries are out of date?

PRs often change tree structure, and sometimes rename nodes - which breaks queries.

My personal opinion is that we are at a point (unless there's some catastrophic parser mistake current grammar has) where we can stabilise and treat breaking changes with a lot more care.

What are you thoughts @ckipp01 and @eed3si9n ?

eed3si9n: > My personal opinion is that we are at a point (unless there's some catastrophic parser mistake current grammar has) where we can stabilise and treat breaking changes with a lot more care.

I agree, except I'd say stabilize. If a PR breaks existing queries we should try working around it by creating a dummy node, or at least know that we are in that territory.

antonsviridov-clearscore: > except I'd say stabilize

image

.github/workflows/ci.yml
keynmol: As a separate point, I think we can extract the Linux workflow as a separate thing, and on Mac/Windows just do `npm run build && npm run test`.

To declutter all those if conditionals.

keynmol (approved):

.github/workflows/ci.yml
```diff
@@ -90,3 +97,23 @@ jobs:
...
+        shell: bash
+        run: cp ./nvim_treesitter/queries/scala/highlights.scm ./queries/highlights.scm
+
+      - name: Check if highlight queries are out of sync with nvim-treesitter
+        if: ${{ runner.os == 'Linux' }}
+        uses: tj-actions/verify-changed-files@v13
+        id: verify-changed-files
+        with:
+          files: |
+            queries/highlights.scm
```
keynmol, line 112: Actually, grammar (but not scanner, I think) changes can also trigger broken highlighting.

ckipp01: Sure, but that will be caught by our own tests. The only reason we are looking here is to see if the highlights are different in our repo than that of the nvim-treesitter repo since this check is after we copied them.

.github/workflows/ci.yml
ckipp01: I can try to follow-up with another pr and clean this up.

Issue #157, opened 2023-01-15
ERROR if type arguments have a `,` at the end


ckipp01:
### Commit of tree-sitter-scala you tested this on

018a6b9d49fb2453495ad131f874f0ef9529b271

### A code sample showing the error

```scala
    val possibleServers: Map[String, Either[
      BuildServerProvider,
      BspConnectionDetails,
    ]] = ???
```

Scalafmt formats a bunch of code like this in the Metals codebase

### Show the error node

```
        val_definition [203, 4] - [206, 12]
          pattern: identifier [203, 8] - [203, 23]
          type: generic_type [203, 25] - [206, 6]
            type: type_identifier [203, 25] - [203, 28]
            type_arguments: type_arguments [203, 28] - [206, 6]
              type_identifier [203, 29] - [203, 35]
              generic_type [203, 37] - [206, 5]
                type: type_identifier [203, 37] - [203, 43]
                type_arguments: type_arguments [203, 43] - [206, 5]
                  type_identifier [204, 6] - [204, 25]
                  type_identifier [205, 6] - [205, 26]
                  ERROR [205, 26] - [206, 4]
```

### What do you expect the tree to look like

I'd expect that error node to just not be there.

### Where are you experiencing this error?

nvim-treesitter

eed3si9n:
Here's my PR for it - https://github.com/tree-sitter/tree-sitter-scala/pull/170

Issue #141, opened 2023-01-13
infix_expression across multiple lines


eed3si9n:
## steps

```scala
def foo: Boolean =
  x
  || a.b(C)
```

## problem

Infix doesn't seem to parse correctly when the operator is on the next line.

```lisp
(compilation_unit [0, 0] - [3, 0]
  (function_definition [0, 0] - [3, 0]
    name: (identifier [0, 4] - [0, 7])
    return_type: (type_identifier [0, 9] - [0, 16])
    body: (indented_block [1, 2] - [3, 0]
      (identifier [1, 2] - [1, 3])
      (infix_expression [2, 2] - [2, 11]
        left: (infix_expression [2, 2] - [2, 7]
          left: (operator_identifier [2, 2] - [2, 4])
          operator: (identifier [2, 5] - [2, 6])
          right: (operator_identifier [2, 6] - [2, 7]))
        operator: (identifier [2, 7] - [2, 8])
        right: (parenthesized_expression [2, 8] - [2, 11]
          (identifier [2, 9] - [2, 10]))))))
```

## expectation

It should work.

```lisp
(compilation_unit [0, 0] - [2, 0]
  (function_definition [0, 0] - [2, 0]
    name: (identifier [0, 4] - [0, 7])
    return_type: (type_identifier [0, 9] - [0, 16])
    body: (indented_block [1, 2] - [2, 0]
      (infix_expression [1, 2] - [1, 13]
        left: (identifier [1, 2] - [1, 3])
        operator: (operator_identifier [1, 4] - [1, 6])
        right: (call_expression [1, 7] - [1, 13]
          function: (field_expression [1, 7] - [1, 10]
            value: (identifier [1, 7] - [1, 8])
            field: (identifier [1, 9] - [1, 10]))
          arguments: (arguments [1, 10] - [1, 13]
            (identifier [1, 11] - [1, 12])))))))
```

## notes

```scala
scala> {
     |   val x = 1
     |     + 1
     | }
val x: Int = 2
```

PR #170, opened 2023-01-19
Trailing commas

eed3si9n:
Fixes https://github.com/tree-sitter/tree-sitter-scala/issues/157

Problem
-------
Trailing commas are not supported.

Solution
--------
This adds support for that.

### DIFF
--- a/script/smoke_test.sh
+++ b/script/smoke_test.sh
@@ -3,8 +3,8 @@
 # This is an integration test to generally check the quality of parsing.
 
 SCALA_SCALA_LIBRARY_EXPECTED=100
-SCALA_SCALA_COMPILER_EXPECTED=67
-DOTTY_COMPILER_EXPECTED=66
+SCALA_SCALA_COMPILER_EXPECTED=66
+DOTTY_COMPILER_EXPECTED=65
 
 if [ ! -d "$SCALA_SCALA_DIR" ]; then
   echo "\$SCALA_SCALA_DIR must be set"
--- a/corpus/definitions.txt
+++ b/corpus/definitions.txt
@@ -174,7 +174,9 @@ Imports
 import PartialFunction.condOpt
 import a.b, c.e
 import reflect.io.{Directory, File, Path}
-
+import a.{
+  b,
+}
 ---
 
 (compilation_unit
@@ -186,7 +188,9 @@ import reflect.io.{Directory, File, Path}
     (identifier) (identifier))
   (import_declaration
     (identifier) (identifier)
-    (namespace_selectors (identifier) (identifier) (identifier))))
+    (namespace_selectors (identifier) (identifier) (identifier)))
+  (import_declaration (identifier)
+    (namespace_selectors (identifier))))
 
 ================================
 Imports: Wildcard
@@ -307,7 +311,10 @@ object O3 extends A {
 Class definitions
 =======================================
 
-class C[T, U] {
+class C[
+  T,
+  U,
+] {
 }
 ---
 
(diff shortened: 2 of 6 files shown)


script/smoke_test.sh
```diff
@@ -3,8 +3,8 @@
 # This is an integration test to generally check the quality of parsing.
 
 SCALA_SCALA_LIBRARY_EXPECTED=100
-SCALA_SCALA_COMPILER_EXPECTED=67
-DOTTY_COMPILER_EXPECTED=66
+SCALA_SCALA_COMPILER_EXPECTED=66
+DOTTY_COMPILER_EXPECTED=65
```
keynmol, line 7: Do you have any idea what regressed here?

eed3si9n:
```
$ diff -U 0 main-reports/report-scala2-compiler.txt report-scala2-compiler.txt | rg '^[+-]\w+'
4:-scala/tools/nsc/ast/parser/TreeBuilder.scala                                 (ERROR [143, 60] - [143, 61])
7:+scala/tools/nsc/backend/jvm/analysis/BackendUtils.scala                      (ERROR [560, 6] - [627, 4])
10:+scala/tools/nsc/tasty/TreeUnpickler.scala                                   (ERROR [368, 10] - [378, 8])
13:+scala/tools/nsc/typechecker/Checkable.scala                                 (ERROR [320, 16] - [326, 13])
17:+scala/tools/nsc/typechecker/splain/SplainFormatting.scala                   (ERROR [360, 4] - [367, 2])
$ diff -U 0 main-reports/report-dotty-compiler.txt report-dotty-compiler.txt | rg '^[+-]\w+'
4:+src/dotty/tools/backend/sjs/JSCodeGen.scala                                                  (ERROR [805, 11] - [805, 12])
7:+src/dotty/tools/dotc/semanticdb/Scala3.scala                                                 (ERROR [163, 4] - [163, 36])
10:+src/dotty/tools/dotc/transform/ExtensionMethods.scala                                       (ERROR [117, 8] - [125, 6])
13:+src/dotty/tools/MainGenericCompiler.scala                                                   (ERROR [175, 21] - [175, 22])
16:+src/dotty/tools/repl/ParseResult.scala                                                      (ERROR [156, 44] - [156, 45])
19:+test-coursier/dotty/tools/coursier/CoursierScalaTests.scala                                 (ERROR [168, 175] - [168, 178])
23:+test/dotty/tools/dotc/SettingsTests.scala                                                   (ERROR [181, 98] - [181, 99])
```

If we take https://github.com/lampepfl/dotty/blob/3.2.1/compiler/src/dotty/tools/backend/sjs/JSCodeGen.scala as an example, the file actually contains a trailing comma: https://github.com/lampepfl/dotty/blob/3.2.1/compiler/src/dotty/tools/backend/sjs/JSCodeGen.scala#L104 so even if it was succeeding previously it was on some false basis. Now it's failing on line 805 due to leading infix (https://github.com/tree-sitter/tree-sitter-scala/issues/141).

Similar story with Scala 2 compiler. https://github.com/scala/scala/blob/v2.13.10/src/compiler/scala/tools/nsc/backend/jvm/analysis/BackendUtils.scala#L280-L288 has trailing commas.

ckipp01: Ah, thanks for the example!

ckipp01 (approved):
Looks good. LGTM!

ckipp01:
Interesting, I was curious what the % would do in Metals with this change since we use trailing commas quite a bit. It was the same, it dropped _just slightly_.

_Before_
`success percentage: 85.33%`

_After_
`success percentage: 85.11%`

PR #174, opened 2023-01-20
refactor(queries): change uppercase from constant to type

ckipp01:
This is in reference to the conversation that was in
https://github.com/tree-sitter/tree-sitter-scala/discussions/168 around
imports and the coloring of the final part. Taken from Eugene's comment,
which made a lot of sense to message

> In Scala we have namespace for terms and types, and they can each
> define the same name, often encouraged as "companion object", so at the
> point of import statement it's ambiguous.

This is relevant in the import case, but I also think I agree with it
most of the time when you have an uppercase identifier. While there may
be times this isn't the case, I think it's a safe default.

If we agree on this, I'll open a pr in nvim-treesitter so that test passes.

### DIFF
--- a/queries/highlights.scm
+++ b/queries/highlights.scm
@@ -241,7 +241,7 @@
 
 (operator_identifier) @operator
 
-((identifier) @constant (#lua-match? @constant "^[A-Z]"))
+((identifier) @type (#lua-match? @type "^[A-Z]"))
 ((identifier) @variable.builtin
  (#lua-match? @variable.builtin "^this$"))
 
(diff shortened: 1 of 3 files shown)


keynmol:
I like it. The colourfulness of the "Before" was really misleading, because it introduced (in most cases) false dichotomy between types, depending on where they appeared.

---
**Before**
Screenshot 2023-01-20 at 16 56 44
**After**
Screenshot 2023-01-20 at 16 59 22

keynmol (approved):

eed3si9n (approved):
Thanks, @ckipp01! LGTM

ckipp01:
Alright https://github.com/nvim-treesitter/nvim-treesitter/pull/4192 got merged in so CI is all green now.

PR #175, opened 2023-01-22
Fix `as` import renaming in Scala 3

keynmol:
In Scala 3, you can rename a single identifier without the curly braces if you use `as`
Doesn't work with `=>`

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -199,6 +199,10 @@ module.exports = grammar({
         choice(
           $.namespace_wildcard,
           $.namespace_selectors,
+          // Only allowed in Scala 3 
+          // ImportExpr        ::= 
+          //    SimpleRef {‘.’ id} ‘.’ ImportSpec |  SimpleRef ‘as’ id
+          $.as_renamed_identifier
         ),
       )),
     )),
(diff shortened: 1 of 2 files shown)


grammar.js
```diff
@@ -215,17 +219,24 @@ module.exports = grammar({
...
+          $.as_renamed_identifier
         )),
       '}'
     ),
 
     // deprecated: Remove when highlight query is updated for Neovim
     _import_selectors: $ => alias($.namespace_selectors, $.import_selectors),
 
-    renamed_identifier: $ => seq(
+    arrow_renamed_identifier: $ => seq(
```
keynmol, line 231: I split the renamed_identifier node as it was a low-tech solution to prevent parsing code that is invalid in Scala 3.

If that's unacceptable, then we can loosen the restriction and use the old renamed_identifier node and just parse incorrect code, if it's for the purspose of highlighting anyways.

ckipp01: My opinion would be that this is probably fine.

ckipp01 (approved):

eed3si9n (approved):
LGTM too

PR #180, opened 2023-01-24
Change directory structure to work well with query-linter module

ghostbuster91:
I suppose that you know https://github.com/nvim-treesitter/playground which helps working with grammar from the neovim. 

Recently I discovered that there is a `query-linter` module that allows you to verify syntax highlighting queries against your grammar in the real time. 

I attach a screenshot from the neovim where the installed version of treesitter is slightly out of date and the linter reports some errors.

 
Screenshot from 2023-01-24 09-09-49

Unfortunately, there is a limitation of the query-linter that requires queries to be put under `./queries/<languaga-name>`. It deviates a little bit from the well defined directory structure but I think that it is not a big deal. 

I added this commit to `git-blame-ignore-revs` not to mess with git history of the file.


ckipp01 (changes requested):
Hey @ghostbuster91, thanks for the pr! I've actually used this a bit when inside of the `nvim-treesitter` repo queries. This is neat! I wish they had a better fallback system, but understandable seeing that it's focused on nvim-treesitter. I see that you set `highlights` inside of the `package.json` so in _theory_ this should be fine for anything actually using the repo for highlights. Although I have no idea if anything else actually is.

I think I'm fine with this, but I'll also wait for another ✅ from one of the other maintainers.

One thing that you'll also want to fix is the CI for checking the queries against that of nvim-treesitter. For example you'll see reference to the old path in a couple places:

https://github.com/tree-sitter/tree-sitter-scala/blob/d67e84cded57f35459bed3f4a2b3db72f715621c/.github/workflows/ci.yml#L104-L118

eed3si9n (approved):
I don't know much about the implications here but LGTM

ghostbuster91:
@ckipp01 sure thing, fixed

.git-blame-ignore-revs
```diff
@@ -0,0 +1,2 @@
+# Move queries to ./queries/scala sub-directory 
+c7e752ae212336b2b8f6eb7d9bc426434c384832
```
ckipp01, line 2: I _think_ this is wrong. Shouldn't this be `889eb65aea8e0f9ce1694ab9ee528df1fa5f75d1`?

ckipp01 (changes requested):

.git-blame-ignore-revs
ghostbuster91: Right, the rebase :+1:

ckipp01 (approved):
Great, thanks @ghostbuster91!

LGTM

PR #190, opened 2023-02-18
Fix compilation error on "cargo build"

smarter:


ckipp01 (approved):
Thanks for the pr @smarter! This is also a pretty good indication that we should have this checked in CI since https://github.com/tree-sitter/tree-sitter-scala/pull/180 should have failed this.

I'll create an issue for this!

PR #195, opened 2023-03-12
fix: Correctly handle _ separator in integer literals

ghostbuster91:
I noticed that syntax highlighting breaks when `_` is present in integer literals. 
image

This PR fixes that.


eed3si9n (approved):
Thanks @ghostbuster91!

PR #196, opened 2023-03-12
fix: Incorrect string interpolation highlighting group

ghostbuster91:
I noticed that values from interpolated strings are not highlighted correctly:

Screenshot from 2023-03-12 22-11-18
 
In order to have correct hl (highlight) group assigned I had to introduce scope tracking feature from the treesitter. After that some hl tests started failing. 

Since some of the changes might be controversial I breakdown them here:
1. in `self: X =>` `self` had a `type` hl group previously. I dunno why. I changed it to `parameter`
2. in `export scanUnit.scan` `scanUnit` was a `namespace` but it is declared few lines above as a variable hence it should has the same hl group as its definition. 

   According to the treesitter documentation:
   > Good syntax highlighting helps the reader to quickly distinguish between the different types of entities in their code. Ideally, if a given entity appears in multiple places, it should be colored the same in each place. 
3. `(interpolation) @none` was too wide and I couldn't get it to work with the rest of the changes, so I narrowed it.

### DIFF
--- a/test/highlight/basics.scala
+++ b/test/highlight/basics.scala
@@ -57,7 +57,7 @@ object Hello {
 // ^ keyword
 //      ^ type
     self: X =>
-//  ^type
+//  ^parameter
 //        ^type
   }
 
@@ -70,8 +70,14 @@ object Hello {
 //               ^keyword
 //                   ^type.definition
 
-  val hello = c"some $stuff"
+  val hello = c"some $mutation ${1}"
 //            ^function.call
 //                   ^punctuation.special
+//                     ^variable
+//                               ^number
+  def meth = ???
+//     ^method
+  val hello2 = c"some $meth"
+//                      ^method
 }
 
--- a/test/highlight/scala3.scala
+++ b/test/highlight/scala3.scala
@@ -141,9 +141,9 @@ class Copier:
 
   export scanUnit.scan
   // ^ include
-  //        ^namespace
+  //        ^variable
   export printUnit.{status as _, *}
   // ^ include
-  //        ^namespace
+  //        ^variable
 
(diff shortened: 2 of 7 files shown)


test/highlight/scala3.scala
```diff
@@ -141,9 +141,9 @@ class Copier:
 
   export scanUnit.scan
   // ^ include
-  //        ^namespace
+  //        ^variable
```
eed3si9n, line 144: Nice.

test/highlight/basics.scala
```diff
@@ -57,7 +57,7 @@ object Hello {
 // ^ keyword
 //      ^ type
     self: X =>
-//  ^type
+//  ^parameter
```
eed3si9n, line 60: This is also great.

eed3si9n (approved):
overall LGTM

CI is failing though:
```:
  ✗ basics.scala
    Failure - row: 58, column: 4, expected highlight 'parameter', actual highlights: 'type'
```

ghostbuster91:
> CI is failing though:
> 
> ```
>   ✗ basics.scala
>     Failure - row: 58, column: 4, expected highlight 'parameter', actual highlights: 'type'
> ```

Indeed, that new query file had to be added in couple different places. Should be fixed now.

ckipp01:
> But the build still fails 🤔

This is because we have a check to make sure our queries match the queries in nvim-treesitter just to warn us if they go out of sync. Since you're introducing more queries here that's sort of expected to fail.  No worries on that. After we merge this I'll sync the queries here with those.

ghostbuster91:
@ckipp01 hmm, I modified slightly the build process and it passed now :D

Issue #198, opened 2023-04-13
Top-level expressions cannot be parsed


keynmol:
### Commit of tree-sitter-scala you tested this on

HEAD

### A code sample showing the error

**test-plugins.sbt**

```scala
addSbtPlugin("com.indoorvivants.vcpkg" % "sbt-vcpkg-native" % "0.0.11")
// hello
```

### Show the error node

Run with `npm exec -- tree-sitter parse test-plugins.sbt`

```
(compilation_unit [0, 0] - [2, 0]
  (ERROR [0, 0] - [2, 0]
    (comment [1, 0] - [1, 8])))
test-plugins.sbt        0 ms    (ERROR [0, 0] - [2, 0])
```

### What do you expect the tree to look like

Correctly parsed

---

### Notes

This is technically not valid Scala, you can't have top level expressions:

```scala
val x = 25
println(x)

@main def hello = ???
```

```
Compiling project (Scala 3.2.0, JVM)
[error] ./test.scala:2:1: Illegal start of toplevel definition
[error] println(x)
[error] ^^^^^^^
Error compiling project (Scala 3.2.0, JVM)
Compilation failed
```

But it's valid in the context of SBT.

@olafurpg patched it in Sourcegraph's fork, but that breaks most tests: https://github.com/sourcegraph/tree-sitter-scala/commit/6cd3cb6074624dd3d88e458343475ca4c701cb06 so we will likely need to rejig other parts of the grammar.

The question of whether we allow it unconditionally is open, but for the purposes of syntax highlighting IMO we should.

bishabosha:
I think given scala script files with scala-cli are now SIP approved, this is certainly more critical

susliko:
I took a look at the errors arising from including `$.expression` into `$._top_level_definition`. It seems, that roughly a half of them is related to the left-associativity of class/object/trait definition (introduced in https://github.com/tree-sitter/tree-sitter-scala/pull/102 in order to miminize memory usage). For example
```scala
class Foo() {}
```
is parsed as
```
(compilation_unit [0, 0] - [2, 0]
  (class_definition [0, 0] - [0, 11]
    name: (identifier [0, 6] - [0, 9])
    class_parameters: (class_parameters [0, 9] - [0, 11]))
  (block [0, 12] - [0, 14]))
```
(`$.block` is now a valid top-level node as a part of `$._simple_expression`)

@eed3si9n fyi

PR #210, opened 2023-05-18
Enable parser on sbt files

ghostbuster91:


ghostbuster91:
> I'd have to play around with it, but I think we need support for top level expressions before we do this no? #198

Ah, right. My sbt file that I was playing around got luckily parsed without almost any errors, though that was just a coincidence.

eed3si9n (approved):

PR #213, opened 2023-05-18
chore: Update nixpkgs to 22.11 and add python3 to devshell

ghostbuster91:
21.05 is already 2 years old. Switching to 22.11 will make sure that binaries are still in the public cache. 

~In addition I added python3 which seems to by needed internally by npm. I didn't notice that before because I had python globally installed.~

I also found out that on the nixos that is more strict regarding the globabl packages tree-sitter installed via npm no longer works. I switched to use tree-sitter from the upstream.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,8 +1,8 @@
 let
   nixpkgs = fetchTarball {
     name = "nixpkgs";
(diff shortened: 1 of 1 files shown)


ckipp01 (commented):
You should be able to rebase on main to fix the CI error.

eed3si9n (approved):
Thanks!

Issue #121, opened 2023-01-10
`given Typeable[T] with ...` doesn't parse


eed3si9n:
I think `given Typeable[T] with ...` doesn't parse - not sure why 
![image](https://user-images.githubusercontent.com/1052965/211519466-af36a3de-d70a-4b51-b793-66e4fedf6133.png)

_Originally posted by @keynmol in https://github.com/tree-sitter/tree-sitter-scala/issues/120#issuecomment-1376998342_

Issue #208, opened 2023-04-29
Anonymous alias givens


susliko:
### Commit of tree-sitter-scala you tested this on

7d348f5

### A code sample showing the error

```scala
object Foo:
  given Ordering[Int] = ???
```

### Show the error node

```
    given_definition [1, 2] - [1, 27]
      return_type: type_identifier [1, 8] - [1, 16]
      ERROR [1, 16] - [1, 21]
        type_parameters [1, 16] - [1, 21]
          name: identifier [1, 17] - [1, 20]
      body: operator_identifier [1, 24] - [1, 27]
```

### What do you expect the tree to look like

Something like 
```
   given_definition [1, 2] - [1, 27]
      type: generic_type [1, 9] - [1, 22]
        type: type_identifier [..] - [..]
        type_arguments: type_arguments [..] - [..]
          type_identifier [..] - [..]
      body: operator_identifier [1, 24] - [1, 27]
```

### Where are you experiencing this error?

nvim-treesitter

susliko:
Probably, a duplicate for https://github.com/tree-sitter/tree-sitter-scala/issues/121

PR #214, opened 2023-05-19
Fix parsing for a few given cases

susliko:
## Summary

Fixes https://github.com/tree-sitter/tree-sitter-scala/issues/121 and https://github.com/tree-sitter/tree-sitter-scala/issues/208

Both structural and alias anonymous given instances are being parsed correctly with this change:
image

There is an extra change: first commit formats `corpus/definitions.txt` with `tree-sitter test -u` of version 0.20.8 (the most recent)

## Explanation

Given instances can have an optional signature:
```
  optional(seq(
    $._function_constructor,
    ':',
  )),
```
E.g. ` A(using B)` in a signature of the given
```scala
given A(using B): C = ???
```
Despite being optional, it was also matched by the tree-sitter for givens without a signature, consuming tokens that should be consumed either by _annotated_type of _structural_instance:
```scala
given C = ???
```
It has something to do with the usage of parsing precedences, though I'm not sure why the GLR was not triggered here (or it was, but did not find path without errors)

### DIFF
--- a/corpus/definitions.txt
+++ b/corpus/definitions.txt
@@ -1,6 +1,6 @@
-=======================================
+================================================================================
 Identifiers
-=======================================
+================================================================================
 
   def m = ???
   def unary_! = true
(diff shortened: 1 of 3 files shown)


eed3si9n (approved):
Nice. Thanks @susliko!

Issue #187, opened 2023-01-31
Generalized type constraints not handled properly


ckipp01:
### Commit of tree-sitter-scala you tested this on

628e0aab6c2f7d31cf3b7d730f964d4fd9b340ee

### A code sample showing the error

```scala
def foo[A]()(using A =:= Int) = ???

def bar[A]()(using A <:< Int) = ???
```

### Show the error node

So the first one with `=:=` doesn't actually an error node, but instead still parses, but incorrectly:

```
function_definition [0, 0] - [0, 35]
  "def" [0, 0] - [0, 3]
  name: identifier [0, 4] - [0, 7]
  type_parameters: type_parameters [0, 7] - [0, 10]
    "[" [0, 7] - [0, 8]
    name: identifier [0, 8] - [0, 9]
    "]" [0, 9] - [0, 10]
  parameters: parameters [0, 10] - [0, 12]
    "(" [0, 10] - [0, 11]
    ")" [0, 11] - [0, 12]
  parameters: parameters [0, 12] - [0, 29]
    "(" [0, 12] - [0, 13]
    "using" [0, 13] - [0, 18]
    parameter [0, 19] - [0, 28]
      name: identifier [0, 19] - [0, 20]
      "=" [0, 21] - [0, 22]
      default_value: postfix_expression [0, 22] - [0, 28]
        operator_identifier [0, 22] - [0, 24]
        identifier [0, 25] - [0, 28]
    ")" [0, 28] - [0, 29]
  "=" [0, 30] - [0, 31]
  body: operator_identifier [0, 32] - [0, 35]
```
`<:<` does however produce an error node:

```
function_definition [2, 0] - [2, 35]
  "def" [2, 0] - [2, 3]
  name: identifier [2, 4] - [2, 7]
  type_parameters: type_parameters [2, 7] - [2, 10]
    "[" [2, 7] - [2, 8]
    name: identifier [2, 8] - [2, 9]
    "]" [2, 9] - [2, 10]
  parameters: parameters [2, 10] - [2, 12]
    "(" [2, 10] - [2, 11]
    ")" [2, 11] - [2, 12]
  parameters: parameters [2, 12] - [2, 29]
    "(" [2, 12] - [2, 13]
    "using" [2, 13] - [2, 18]
    parameter [2, 19] - [2, 20]
      name: identifier [2, 19] - [2, 20]
    ERROR [2, 21] - [2, 28]
      "<:" [2, 21] - [2, 23]
[...]

PR #215, opened 2023-05-21
fix: Anonymous parameters in `using` clauses

susliko:
Problem
-------
`$.parameters` required each param to have a name, so `using` clauses like `def x(using X =:= X)` failed to parse

Solution
-------
Introduce `$._using_parameters_clause`, which allows anonymous parameters

Resolves https://github.com/tree-sitter/tree-sitter-scala/issues/187

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -620,10 +620,33 @@ module.exports = grammar({
       ')'
     )),
 
-    parameters: $ => seq(
+    /*
+    * DefParamClauses   ::=  {DefParamClause} [[nl] ‘(’ [‘implicit’] DefParams ‘)’]
+    * DefParamClause    ::=  [nl] ‘(’ DefParams ‘)’ | UsingParamClause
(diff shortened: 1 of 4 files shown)


grammar.js
```diff
@@ -637,11 +660,16 @@ module.exports = grammar({
...
+    * DefParam          ::=  {Annotation} [‘inline’] Param
+    * Param             ::=  id ‘:’ ParamType [‘=’ Expr]
+    */
     parameter: $ => seq(
       repeat($.annotation),
       optional($.inline_modifier),
       field('name', $._identifier),
-      optional(seq(':', field('type', $._param_type))),
+      ':',
+      field('type', $._param_type),
```
susliko, line 672: Type annotation for parameters shouldn't be optional

eed3si9n (approved):
Thanks @susliko!

Issue #188, opened 2023-02-06
Incorrect parse for if_expression with extra line between alternative


rochala:
### Commit of tree-sitter-scala you tested this on

latest

### A code sample showing the error

<img width="243" alt="image" src="https://user-images.githubusercontent.com/48657087/216967112-9c0efca9-f4dd-41a2-97c2-0a6c91eab511.png">

```scala
  if true then
    println()

  else
    ???

  if true then
    println()
  else
    println()
```

### Show the error node

``` 
    if_expression [211, 2] - [214, 2]
      condition: boolean_literal [211, 5] - [211, 9]
      consequence: indented_block [212, 4] - [214, 2]
        call_expression [212, 4] - [212, 13]
          function: identifier [212, 4] - [212, 11]
          arguments: arguments [212, 11] - [212, 13]
    identifier [214, 2] - [214, 6]
    operator_identifier [215, 4] - [215, 7]
    if_expression [217, 2] - [222, 2]
      condition: boolean_literal [217, 5] - [217, 9]
      consequence: indented_block [218, 4] - [219, 2]
        call_expression [218, 4] - [218, 13]
          function: identifier [218, 4] - [218, 11]
          arguments: arguments [218, 11] - [218, 13]
      alternative: indented_block [220, 4] - [222, 2]
        call_expression [220, 4] - [220, 13]
          function: identifier [220, 4] - [220, 11]
          arguments: arguments [220, 11] - [220, 13]

```

### What do you expect the tree to look like

`if_expression` should contain `alternative` even if there additional line between them and they are intended the same.

```     
    if_expression [211, 2] - [216, 2]
      condition: boolean_literal [211, 5] - [211, 9]
      consequence: indented_block [212, 4] - [213, 2]
        call_expression [212, 4] - [212, 13]
          function: identifier [212, 4] - [212, 11]
          arguments: arguments [212, 11] - [212, 13]
      alternative: indented_block [214, 4] - [216, 2]
        operator_identifier [214, 4] - [214, 7]
    if_expression [216, 2] - [221, 2]
      condition: boolean_literal [216, 5] - [216, 9]
      consequence: indented_block [217, 4] - [218, 2]
        call_expression [217, 4] - [217, 13]
          function: identifier [217, 4] - [217, 11]
          arguments: arguments [217, 11] - [217, 13]
      alternative: indented_block [219, 4] - [221, 2]
        call_expression [219, 4] - [219, 13]
[...]

Issue #199, opened 2023-04-18
Comments in method chains


domaspoliakas:
### Commit of tree-sitter-scala you tested this on

7d348f51e442563f4ab2b6c3e136dac658649f93

### A code sample showing the error

```scala
def potato = 
  "hek"
    .tail
    // Hello
    .tail
    .tail
```

### Show the error node

```
function_definition [0, 0] - [6, 0]
  name: identifier [0, 4] - [0, 10]
  ERROR [1, 2] - [3, 4]
    field_expression [1, 2] - [2, 9]
      value: string [1, 2] - [1, 7]
      field: identifier [2, 5] - [2, 9]
  comment [3, 4] - [3, 12]
  body: indented_block [4, 4] - [6, 0]
    ERROR [4, 4] - [4, 5]
    field_expression [4, 5] - [5, 9]
      value: identifier [4, 5] - [4, 9]
      field: identifier [5, 5] - [5, 9]
```

### What do you expect the tree to look like

This is what it looks like if I take the comment out
```
function_definition [0, 0] - [5, 0]
  name: identifier [0, 4] - [0, 10]
  body: indented_block [1, 2] - [5, 0]
    field_expression [1, 2] - [4, 9]
      value: field_expression [1, 2] - [3, 9]
        value: field_expression [1, 2] - [2, 9]
          value: string [1, 2] - [1, 7]
          field: identifier [2, 5] - [2, 9]
        field: identifier [3, 5] - [3, 9]
      field: identifier [4, 5] - [4, 9]

```

### Where are you experiencing this error?

nvim-treesitter

ghostbuster91:
I gave it a shot but I was unable to make the method chain with comment in the middle to parse correctly. 

Putting optional comment in the `field_expression` doesn't work. 

I found that it works for java parser https://github.com/tree-sitter/tree-sitter-java/blob/c194ee5e6ede5f26cf4799feead4a8f165dcf14d/grammar.js#L33 
I tried to modify scala grammar based on the java one but the issue remained.

susliko:
This issue can be marked resolved as https://github.com/tree-sitter/tree-sitter-scala/pull/217 was merged

ckipp01:
Closed via #217 

Thanks @susliko!

PR #217, opened 2023-05-22
fix: Handle newlines and comments in the middle of expressions

susliko:
Resolves https://github.com/tree-sitter/tree-sitter-scala/issues/188, https://github.com/tree-sitter/tree-sitter-scala/issues/199

Problem:
-------
Code like
```scala
if true ()

else ()
```
or
```scala
a
 .b
 // comment
 .c
```
is not parsed correctly, for extra `$._automated_semicolon` tokens,
which are produced after newlines

Solution:
-------
Tweak `scanner.c` to stop producing AUTOMATED_SEMICOLON tokens when
encountering comments, else/catch clauses, etc.

Extra changes:
-------
corpus/expressions.txt formatted with `tree-sitter test -u` in the first commit

Note:
-------
The following code still won't be parsed
```scala
a
  /* comment */ .b
```
as I haven't figured out a way both to parse it and code like
```scala
val a = 1
/* comment */ val b = 2
```
where AUTOMATED_SEMICOLON is needed to distinguish consecutive blocks

### DIFF
--- a/src/scanner.c
+++ b/src/scanner.c
@@ -43,6 +43,7 @@ void tree_sitter_scala_external_scanner_deserialize(void *p, const char *b,
 }
 
 static void advance(TSLexer *lexer) { lexer->advance(lexer, false); }
+static void skip(TSLexer *lexer) { lexer->advance(lexer, true); }
 
 static bool scan_string_content(TSLexer *lexer, bool is_multiline, bool has_interpolation) {
   unsigned closing_quote_count = 0;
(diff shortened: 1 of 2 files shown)


susliko:
@eed3si9n this one is ready for review
 It would be great if you could take a look as you have an experience with the external parser

src/scanner.c
```diff
@@ -166,14 +167,49 @@ bool tree_sitter_scala_external_scanner_scan(void *payload, TSLexer *lexer,
...
-    // expression, in which case we don't recognize it as an automatic semicolon.
-    if (lexer->lookahead == '.') return false;
+    // AUTOMATIC_SEMICOLON should not be issued in the middle of expressions
+    // Thus, we exit this branch when encountering comments, else/catch clauses, etc.
 
     lexer->mark_end(lexer);
     lexer->result_symbol = AUTOMATIC_SEMICOLON;
 
-    if (newline_count > 1) return true;
+    // Probably, a multi-line field expression, e.g.
```
eed3si9n, line 176: Thanks for this comment here.

eed3si9n (approved):
Overall LGTM

Issue #200, opened 2023-04-22
Pattern matching with guards


susliko:
### Commit of tree-sitter-scala you tested this on

7d348f5

### A code sample showing the error

```scala
object PatMatWithGuards {
  val y = 1
  val x = 1
  x match {
    case x if x == y => x
    case _ => x
  }
}
```
That's one of the cases when pattern matching with guards is not parsed

### Show the error node

Match-block parsed as
```
    match_expression [3, 2] - [6, 3]
      value: identifier [3, 2] - [3, 3]
      body: case_block [3, 10] - [6, 3]
        ERROR [4, 4] - [6, 2]
          identifier [4, 9] - [4, 10]
          identifier [4, 14] - [4, 15]
          operator_identifier [4, 16] - [4, 18]
          lambda_expression [4, 19] - [6, 2]
            parameters: identifier [4, 19] - [4, 20]
            identifier [4, 24] - [4, 25]
            ERROR [5, 4] - [5, 8]
            lambda_expression [5, 9] - [5, 15]
              parameters: wildcard [5, 9] - [5, 10]
              identifier [5, 14] - [5, 15]

```
`y => x` is thought to be a lambda expression

### What do you expect the tree to look like

```
    match_expression [3, 2] - [6, 3]
      value: identifier [3, 2] - [3, 3]
      body: case_block [3, 10] - [6, 3]
        case_clause [4, 4] - [5, 4]
          pattern: identifier [4, 9] - [4, 10]
          guard [4, 11] - [4, 23]
            condition: infix_expression [4, 14] - [4, 23]
              left: identifier [4, 14] - [4, 15]
              operator: operator_identifier [4, 16] - [4, 18]
              right: identifier [4, 19] - [4, 20]
          body: identifier [4, 24] - [4, 25]
        case_clause [5, 4] - [6, 2]
          pattern: wildcard [5, 9] - [5, 10]
          body: identifier [5, 14] - [5, 15]
```

### Where are you experiencing this error?

nvim-treesitter

PR #221, opened 2023-05-24
Fix case clauses with guards

susliko:
Fixes https://github.com/tree-sitter/tree-sitter-scala/issues/200

Problem
-------
Guards in case clauses are thought to be lambda expressions sometimes, e.g. in
```scala
{case a if b => c}
```
This happens, because `$.lambda_expression` has a higher parsing precedence than `$._simple_expression`

Solution
-------
Avoid using numeric parsing precedence for `$.lambda_expression`. Resolve specific conflicts instead

### DIFF
--- a/script/smoke_test.sh
+++ b/script/smoke_test.sh
@@ -3,8 +3,8 @@
 # This is an integration test to generally check the quality of parsing.
 
 SCALA_SCALA_LIBRARY_EXPECTED=100
-SCALA_SCALA_COMPILER_EXPECTED=68
-DOTTY_COMPILER_EXPECTED=66
+SCALA_SCALA_COMPILER_EXPECTED=84
+DOTTY_COMPILER_EXPECTED=71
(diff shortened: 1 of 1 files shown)


susliko:
This change bumps parsing reports for compiler codebases:
```
scala: 68 -> 84
dotty: 66 -> 71
```

eed3si9n (approved):
Thanks @susliko!

eed3si9n (approved):
Thanks @susliko!

keynmol:
Those smoke test bumps are great!

Note, @susliko, that you don't have to check in `parser.c` or `grammar.json`: https://github.com/tree-sitter/tree-sitter-scala/blob/master/CONTRIBUTING.md#generation

Just the grammar and the test changes.

susliko:
@keynmol 
Oh, nice! I've missed this automation for parser generation

Issue #202, opened 2023-04-22
`derives` clause


susliko:
### Commit of tree-sitter-scala you tested this on

7d348f5

### A code sample showing the error

```scala
object Derivations {
  case class X() derives A
}
```

### Show the error node

```
    class_definition [1, 2] - [1, 16]
      name: identifier [1, 13] - [1, 14]
      class_parameters: class_parameters [1, 14] - [1, 16]
    ERROR [1, 17] - [1, 26]
```

### What do you expect the tree to look like

```
    class_definition [1, 2] - [1, 16]
      name: identifier [1, 13] - [1, 14]
      class_parameters: class_parameters [1, 14] - [1, 16]
      TODO_DERIVES_CLAUSE [1, 17] - [1, 26]
```

### Where are you experiencing this error?

nvim-treesitter

PR #222, opened 2023-05-24
Derives clause for classes and objects

susliko:
Closes https://github.com/tree-sitter/tree-sitter-scala/issues/202

Problem
-------
Derives clause is only defined for enums, but not for classes and objects

Solution
-------
Include optional `$.derives_clause` in class and object definitions

Concerns
-------
This change increases `tree-sitter gen` execution time by ~50 percent (while not growing the size of  `parser.c` significantly). This slowdown is similar to to the one from optional branches of  else-clause in `if_expression`, catch-clause of in `try_expression`, etc.

I've opened a discussion on this topic some time ago, but no progress there https://github.com/tree-sitter/tree-sitter/discussions/1948

### DIFF
--- a/corpus/definitions.txt
+++ b/corpus/definitions.txt
@@ -421,6 +421,22 @@ object O3 extends A {
     (template_body)))
 
 ================================================================================
+Object definitions
+================================================================================
+case object A extends B derives C, D {}
+--------------------------------------------------------------------------------
+
+(compilation_unit
+  (object_definition
+    (identifier)
+    (extends_clause
+      (type_identifier))
+    (derives_clause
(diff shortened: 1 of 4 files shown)


eed3si9n:
Responded to the discussion - https://github.com/tree-sitter/tree-sitter/discussions/1948#discussioncomment-5993481

susliko:
Okay, introduction of `$._class_definition` rule, that matches everything after `'class'` keyword (similarly to `$._object_definition`) helped to restrain the states and memory usage blow 😮

eed3si9n (approved):
Thanks @susliko!

PR #231, opened 2023-05-24
Add complexity check to smoke_test.sh

sideeffffect:
@eed3si9n @keynmol @ckipp01


eed3si9n (approved):
Thanks @sideeffffect!

PR #232, opened 2023-05-24
smoke_test.sh: attribute the definition complexity to proper line

sideeffffect:


eed3si9n (approved):
Thanks @sideeffffect!

PR #234, opened 2023-05-25
SPM support

mattmassicotte:
Hi friends! Me again :)

This adds Swift bindings and Swift Package Manager (SPM) support. This should require no maintenance on your part, and should not impact the actual parsers in any way. Other similar changes were made here:

https://github.com/tree-sitter/tree-sitter-go
https://github.com/tree-sitter/tree-sitter-c
https://github.com/tree-sitter/tree-sitter-haskell

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 node_modules
 build
+/.build
 target
 *.log
 test.scala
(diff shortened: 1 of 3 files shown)


keynmol:
I don't fully understand it but that never stopped me from merging code!

Thanks :)

Issue #223, opened 2023-05-24
Using clause in call expression


susliko:
### Commit of tree-sitter-scala you tested this on

9cb4266

### A code sample showing the error

```scala
object X:
  f(using A.B)
```

### Show the error node

```
(compilation_unit [0, 0] - [2, 0]
  (object_definition [0, 0] - [2, 0]
    name: (identifier [0, 7] - [0, 8])
    body: (template_body [0, 8] - [2, 0]
      (call_expression [1, 2] - [1, 14]
        function: (identifier [1, 2] - [1, 3])
        arguments: (arguments [1, 3] - [1, 14]
          (field_expression [1, 4] - [1, 13]
            value: (identifier [1, 4] - [1, 9])
            (ERROR [1, 10] - [1, 11]
              (identifier [1, 10] - [1, 11]))
            field: (identifier [1, 12] - [1, 13])))))))
```

### What do you expect the tree to look like

With a `using clause`, like in function definitions

### Where are you experiencing this error?

nvim-treesitter

PR #235, opened 2023-05-25
Using clause in $.arguments

susliko:
Resolves https://github.com/tree-sitter/tree-sitter-scala/issues/223

Problem
-------
Scala 3 introduced using clauses to function call expressions:
```scala
f(using A)
```
The grammar did not allow that

Solution
-------
Add `choice` between regular and using-clause argument lists in `$.arguments`


eed3si9n (approved):
Thanks @susliko!

Issue #227, opened 2023-05-24
Soft modifiers as function names in call expressions


susliko:
### Commit of tree-sitter-scala you tested this on

9cb4266

### A code sample showing the error

```scala
object X:
  def open() = ???
  open()
```

### Show the error node

```
object_definition [0, 0] - [4, 0]
  name: identifier [0, 7] - [0, 8]
  body: template_body [0, 8] - [4, 0]
    function_definition [1, 2] - [1, 18]
      name: identifier [1, 6] - [1, 10]
      parameters: parameters [1, 10] - [1, 12]
      body: operator_identifier [1, 15] - [1, 18]
    ERROR [2, 2] - [2, 6]
      open_modifier [2, 2] - [2, 6]
    unit [2, 6] - [2, 8]
```

### What do you expect the tree to look like

```
object_definition [0, 0] - [4, 0]
  name: identifier [0, 7] - [0, 8]
  body: template_body [0, 8] - [4, 0]
    function_definition [1, 2] - [1, 18]
      name: identifier [1, 6] - [1, 10]
      parameters: parameters [1, 10] - [1, 12]
      body: operator_identifier [1, 15] - [1, 18]
    call_expression [2, 2] - [2, 8]
      function: identifier [2, 2] - [2, 6]
      arguments: arguments [2, 6] - [2, 8]
```
[List](https://docs.scala-lang.org/scala3/reference/soft-modifier.html) of soft modifiers

PR #237, opened 2023-05-25
Allow soft keywords as identifiers

susliko:
Resolves https://github.com/tree-sitter/tree-sitter-scala/issues/227

Problem
-------
Soft keywords ('open', 'end', 'infix', etc.) are not allowed to be used as identifiers, e.g. in
```scala
open()
```
Solution
-------
Introduce `$._soft_identifier` with a lower precedence than modifiers


susliko:
This one is ready for review

cc @eed3si9n

grammar.js
```diff
@@ -63,6 +62,10 @@ module.exports = grammar({
     $.literal,
   ],
 
+  precedences: $ => [
```
eed3si9n, line 65: Interesting. I wasn't aware of this feature. Could you add a comment linking to a doc (maybe it's just https://github.com/tree-sitter/tree-sitter/pull/939 for now?) for this?

susliko: Sure!
These relative precedences seem to bring more transparency compared to global numeric precedences

eed3si9n (approved):
Thanks @susliko!

Issue #225, opened 2023-05-24
Quote as last character of multi-line string


susliko:
### Commit of tree-sitter-scala you tested this on

9cb4266

### A code sample showing the error

```scala
object X:
  val x = """""""
```

### Show the error node

```
(compilation_unit [0, 0] - [2, 0]
  (object_definition [0, 0] - [2, 0]
    name: (identifier [0, 7] - [0, 8])
    body: (template_body [0, 8] - [2, 0]
      (val_definition [1, 2] - [1, 17]
        pattern: (identifier [1, 6] - [1, 7])
        value: (string [1, 10] - [1, 17]))
      (ERROR [1, 17] - [1, 18]))))
```

### What do you expect the tree to look like

```
(compilation_unit [0, 0] - [2, 0]
  (object_definition [0, 0] - [2, 0]
    name: (identifier [0, 7] - [0, 8])
    body: (template_body [0, 8] - [2, 0]
      (val_definition [1, 2] - [1, 17]
        pattern: (identifier [1, 6] - [1, 7])
        value: (string [1, 10] - [1, 17])))))
```

PR #239, opened 2023-05-25
Allow quotes to be last characters of multi-line string

susliko:
Resolves https://github.com/tree-sitter/tree-sitter-scala/issues/225

Problem
-------
Strings like `"""""""` are valid in scala, but not in the grammar

Solution
-------
In `scanner.c` emit multi-line string only when at least 3 closing quotes found and the next lookahead is not a quote

Also, corpus/literals.txt was formatted with `tree-sitter test -u`

### DIFF
--- a/src/scanner.c
+++ b/src/scanner.c
@@ -55,7 +55,7 @@ static bool scan_string_content(TSLexer *lexer, bool is_multiline, bool has_inte
         lexer->result_symbol = has_interpolation ? INTERPOLATED_STRING_END : SIMPLE_STRING;
         return true;
       }
-      if (closing_quote_count == 3) {
+      if (closing_quote_count >= 3 && lexer->lookahead != '"') {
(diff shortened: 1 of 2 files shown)


src/scanner.c
```diff
@@ -55,10 +55,10 @@ static bool scan_string_content(TSLexer *lexer, bool is_multiline, bool has_inte
         lexer->result_symbol = has_interpolation ? INTERPOLATED_STRING_END : SIMPLE_STRING;
         return true;
       }
-      if (closing_quote_count == 3) {
+      if (closing_quote_count >= 3 && lexer->lookahead != '"') {
         lexer->result_symbol = has_interpolation ? INTERPOLATED_MULTILINE_STRING_END : SIMPLE_MULTILINE_STRING;
         return true;
-      }
+      }
```
eed3si9n, line 61: nit: There's extra whitespace.

eed3si9n (approved):
LGTM, with minor comment

Issue #205, opened 2023-04-22
Match types


susliko:
### Commit of tree-sitter-scala you tested this on

7d348f5

### A code sample showing the error

```scala
object MatchTypes {
  type Elem[X] = X match
    case String => Char
}
```

### Show the error node

```
    ERROR [1, 2] - [2, 8]
      type_definition [1, 2] - [1, 18]
        name: type_identifier [1, 7] - [1, 11]
        type_parameters: type_parameters [1, 11] - [1, 14]
          name: identifier [1, 12] - [1, 13]
        type: type_identifier [1, 17] - [1, 18]
    self_type [2, 9] - [2, 18]
      identifier [2, 9] - [2, 15]
    identifier [2, 19] - [2, 23]
```

### What do you expect the tree to look like

I'd expect `match` in type definition to be handled 

### Where are you experiencing this error?

nvim-treesitter

eed3si9n:
Here's my PR for this - https://github.com/tree-sitter/tree-sitter-scala/pull/244

PR #244, opened 2023-05-26
Match type

eed3si9n:
Fixes https://github.com/tree-sitter/tree-sitter-scala/issues/205

This implements support for Scala 3 match types.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -715,6 +715,12 @@ module.exports = grammar({
       $._outdent,
     )),
 
+    _indented_type_cases: $ => prec.left(PREC.end_marker, seq(
+      $._indent,
+      repeat1($.type_case_clause),
+      $._outdent,
+    )),
+
     // ---------------------------------------------------------------
     // Types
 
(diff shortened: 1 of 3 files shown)


test/highlight/scala3.scala
```diff
@@ -23,6 +23,14 @@ object O1:
   //^keyword.function
   //  ^method
 
+object O2:
+  type Elem[A] = A match
+  //^keyword
+  //   ^type.definition
+    case String   => Char
```
keynmol, line 30: Do the actual types (both on left and right of =>) have highlighting on them? They should, so may be it's worth adding them to the test explicitly.

grammar.js
```diff
@@ -790,8 +804,24 @@ module.exports = grammar({
       field('selector', $._type_identifier),
     ),
 
-    function_type: $ => prec.right(seq(
+    match_type: $ => prec.left(PREC.control, seq(
```
susliko, line 807: Is numeric precedence necessary here? I think, we should minimize its usage as it has a very opaque effect on the result of parsing. 
Maybe, relative precedences (opposed to global) would be a better option

eed3si9n: Good call. At some point this was conflicting with other types, but it seems like I can remove it now.

eed3si9n: Addressed in 41dcb78.

ckipp01 (approved):
Nice! LGTM!

Issue #224, opened 2023-05-24
Braceless try-catch on the same line


susliko:
### Commit of tree-sitter-scala you tested this on

9cb4266

### A code sample showing the error

```scala
object X:
  try sample catch case e: Throwable => x
```

### Show the error node

```
(compilation_unit [0, 0] - [2, 0]
  (object_definition [0, 0] - [2, 0]
    name: (identifier [0, 7] - [0, 8])
    body: (template_body [0, 8] - [2, 0]
      (try_expression [1, 2] - [1, 25]
        body: (identifier [1, 6] - [1, 12])
        (catch_clause [1, 13] - [1, 25]
          (postfix_expression [1, 19] - [1, 25]
            (identifier [1, 19] - [1, 23])
            (identifier [1, 24] - [1, 25]))))
      (ERROR [1, 25] - [1, 41]
        (identifier [1, 27] - [1, 36])))))
```

### What do you expect the tree to look like

```
(compilation_unit [0, 0] - [5, 0]
  (object_definition [0, 0] - [5, 0]
    name: (identifier [0, 7] - [0, 8])
    body: (template_body [0, 8] - [5, 0]
      (try_expression [1, 2] - [5, 0]
        body: (indented_block [2, 4] - [3, 2]
          (identifier [2, 4] - [2, 10]))
        (catch_clause [3, 2] - [5, 0]
          (indented_cases [4, 4] - [5, 0]
            (case_clause [4, 4] - [4, 26]
              pattern: (typed_pattern [4, 9] - [4, 21]
                pattern: (identifier [4, 9] - [4, 10])
                type: (type_identifier [4, 12] - [4, 21]))
              body: (identifier [4, 25] - [4, 26]))))))))
```

### Where are you experiencing this error?

nvim-treesitter

PR #245, opened 2023-05-26
Better catch clause

susliko:
Resolves https://github.com/tree-sitter/tree-sitter-scala/issues/224

Problem
-------
Catch-clause is not parsed when case is on the same line as 'catch':
```scala
try a catch b: C => d
```

Solution
-------
Intoduce `$._expr_case_clause` alternative to the catch-clause


eed3si9n (approved):
Thanks @susliko!

Issue #203, opened 2023-04-22
Type bounds, when upper specified after lower


susliko:
### Commit of tree-sitter-scala you tested this on

7d348f5

### A code sample showing the error

```scala
object LowerUpperBounds {
  type T >: L <: U
}
```

### Show the error node

```
    type_definition [1, 2] - [1, 18]
      name: type_identifier [1, 7] - [1, 8]
      bound: lower_bound [1, 9] - [1, 18]
        type: infix_type [1, 12] - [1, 18]
          left: type_identifier [1, 12] - [1, 13]
          operator: operator_identifier [1, 14] - [1, 16]
          right: type_identifier [1, 17] - [1, 18]
```
Upper bound is detected as an infix type

The case when lower bound is specified after upper is parsed as expected:
```scala
  type T <: U >: L
```
```
    type_definition [1, 2] - [1, 18]
      name: type_identifier [1, 7] - [1, 8]
      bound: upper_bound [1, 9] - [1, 13]
        type: type_identifier [1, 12] - [1, 13]
      bound: lower_bound [1, 14] - [1, 18]
        type: type_identifier [1, 17] - [1, 18]
```

### What do you expect the tree to look like

```
    type_definition [1, 2] - [1, 18]
      name: type_identifier [1, 7] - [1, 8]
      bound: lower_bound [1, 9] - [1, 13]
        type: type_identifier [1, 12] - [1, 13]
      bound: upper_bound [1, 14] - [1, 18]
        type: type_identifier [1, 17] - [1, 18]
```

### Where are you experiencing this error?

nvim-treesitter

PR #247, opened 2023-05-27
Correct type bounds

susliko:
Fixes https://github.com/tree-sitter/tree-sitter-scala/issues/203

Problem
-------
The grammar assumes upper type bound to go before lower type bound.
Although, in Scala it's the opposite:
```
TypeBounds ::= [‘>:’ Type] [‘<:’ Type]
```

Solution
-------
Swap type bounds' order in the grammar

Also, `corpus/types.txt` is formatted with `tree-sitter test -u`

### DIFF
--- a/corpus/types.txt
+++ b/corpus/types.txt
@@ -302,6 +302,24 @@ class A[B >: C]
         (type_identifier)))))
 
 ================================================================================
+Lower and upper bounds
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Thanks @susliko!

Issue #201, opened 2023-04-22
Annotations on enums


susliko:
### Commit of tree-sitter-scala you tested this on

7d348f5

### A code sample showing the error

```scala
object AnnotsWithEnums {
  @Annot
  enum Enum:
    @Annot case A
}
```

### Show the error node

```
    ERROR [1, 2] - [1, 8]
      annotation [1, 2] - [1, 8]
        name: type_identifier [1, 3] - [1, 8]
    enum_definition [2, 2] - [4, 0]
      name: identifier [2, 7] - [2, 11]
      body: enum_body [2, 11] - [4, 0]
        ERROR [3, 4] - [3, 15]
          annotation [3, 4] - [3, 10]
            name: type_identifier [3, 5] - [3, 10]
        identifier [3, 16] - [3, 17]
```

### What do you expect the tree to look like

```
    enum_definition [1, 2] - [2, 16]
      annotation [1, 2] - [1, 8]
        name: type_identifier [1, 3] - [1, 8]
      name: identifier [2, 13] - [2, 14]
      body: enum_body [2, 11] - [4, 0]
        enum_case_definitions [3, 4] - [3, 10]
          simple_enum_case [3, 9] - [3, 10]
            annotation [..] - [..]
              name: type_identifier [..] - [..]
            name: identifier [3, 9] - [3, 10]
```

### Where are you experiencing this error?

nvim-treesitter

PR #248, opened 2023-05-27
Annotations on enums

susliko:
Fixes https://github.com/tree-sitter/tree-sitter-scala/issues/201

Problem
-------
Annotations on Scala 3 enums and their cases are not supported, e.g.
```scala
@A enum E:
  @A("") @B case A

```

Solution
-------
Allow annotations in enum definitions

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -113,6 +113,7 @@ module.exports = grammar({
     ),
 
     enum_definition: $ => seq(
+      repeat($.annotation),
       'enum',
       $._class_constructor,
       field('extend', optional($.extends_clause)),
@@ -148,6 +149,7 @@ module.exports = grammar({
     ),
 
     enum_case_definitions: $ => seq(
+      repeat($.annotation),
       'case',
       choice(
         commaSep1($.simple_enum_case),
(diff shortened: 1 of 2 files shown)


grammar.js
```diff
@@ -146,7 +147,8 @@ module.exports = grammar({
       )
     ),
 
-    enum_case_definitions: $ => seq(
+    enum_case_definition: $ => seq(
```
eed3si9n, line 150: Why was this renamed here? Since it's a `seq(...)` wouldn't it make sense to keep it plural?

susliko: Hm
I read it as a "definition of a single enum case", hence i'd prefer it singular 🤔 
One might think that this rule handles definitions of all enum cases
EBNF calls it just EnumCase

eed3si9n: These names leak out to be tree-sitter queries in various editors, so unless there's a good reason we should avoid renaming them, and even if we do try to make an alias to keep backward compatibility.

susliko: Reverted the rename

eed3si9n (approved):
Thanks 🙏!

Issue #228, opened 2023-05-24
By-name arguments in class constructors


susliko:
### Commit of tree-sitter-scala you tested this on

9cb4266

### A code sample showing the error

```scala
class X (x: => T) 
```

### Show the error node

```
(compilation_unit [0, 0] - [1, 0]
  (class_definition [0, 0] - [0, 17]
    name: (identifier [0, 6] - [0, 7])
    class_parameters: (class_parameters [0, 8] - [0, 17]
      (class_parameter [0, 9] - [0, 14]
        name: (identifier [0, 9] - [0, 10])
        type: (type_identifier [0, 12] - [0, 14]))
      (ERROR [0, 15] - [0, 16]
        (identifier [0, 15] - [0, 16])))))
```

### What do you expect the tree to look like

```
  (class_definition [0, 0] - [0, 17]
    name: (identifier [0, 6] - [0, 7])
    class_parameters: (class_parameters [0, 8] - [0, 17]
      (class_parameter [0, 9] - [0, 14]
        name: (identifier [0, 9] - [0, 10])
        type: (lazy_parameter_type [..] - [..]
          type: (type_identifier [..] - [..]))))))
```

PR #249, opened 2023-05-27
By-name class parameters

susliko:
Fixes https://github.com/tree-sitter/tree-sitter-scala/issues/228

Problem
-------
By-name parameters in class definitions were not supported:
```scala
class A(a: => B)
```

Solution
-------
Support by-name parameters by changing the rule describing class parameter type from `$._type` to `$._param_type`


eed3si9n (approved):
Thanks @susliko!

Issue #230, opened 2023-05-24
Self-type in braceless definitions


susliko:
### Commit of tree-sitter-scala you tested this on

9cb4266

### A code sample showing the error

```scala
trait A:
  this: Int =>
```

### Show the error node

```
(compilation_unit [0, 0] - [3, 0]
  (trait_definition [0, 0] - [0, 7]
    name: (identifier [0, 6] - [0, 7]))
  (ERROR [0, 7] - [1, 14]
    (identifier [1, 2] - [1, 6])
    (identifier [1, 8] - [1, 11])))
```

### What do you expect the tree to look like

```
(compilation_unit [0, 0] - [4, 0]
  (trait_definition [0, 0] - [2, 1]
    name: (identifier [0, 6] - [0, 7])
    body: (template_body [0, 8] - [2, 1]
      (self_type [1, 2] - [1, 14]
        (identifier [1, 2] - [1, 6])
        (type_identifier [1, 8] - [1, 11])))))
```

PR #250, opened 2023-05-27
Self-types in braceless trait definitions

susliko:
Fixes https://github.com/tree-sitter/tree-sitter-scala/issues/230

Problem
-------
Self-types are not parsed in braceless trait definitions:
```scala
trait A:
  this: B =>
````

Solution
-------
Move existing `$.self_type` rule in trait definition into indented scope.
Rebalance `$.self_type` precedence.
Without explicit precedence it looses to `$.call_expression` in
```scala
trait A {
  self: B =>
  def f = ""
}
```
With static precedence instead of dynamic it leads to errors in
definitions like:
```scala
class A
  x: Int // this should be $.ascription_expression
enum B:
  case C
```

### DIFF
--- a/corpus/definitions.txt
+++ b/corpus/definitions.txt
@@ -1584,10 +1584,11 @@ end Test
         (integer_literal)))))
 
 ================================================================================
-Transparent traits (Scala 3)
+Traits (Scala 3)
 ================================================================================
 
 transparent trait Kind:
+  this: A =>
   def test = 1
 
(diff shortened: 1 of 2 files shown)


susliko:
This PR is ready for review!

cc @eed3si9n

eed3si9n (approved):
Thanks @susliko!

eed3si9n (approved):
Thanks @susliko!

Issue #226, opened 2023-05-24
Nested block comments


susliko:
### Commit of tree-sitter-scala you tested this on

9cb4266

### A code sample showing the error

```scala
/* /* */ */
```

### Show the error node

```
(compilation_unit [0, 0] - [1, 0]
  (comment [0, 0] - [0, 8])
  (ERROR [0, 9] - [1, 0]
    (ERROR [0, 10] - [0, 11])))
```

### What do you expect the tree to look like

```
(compilation_unit [0, 0] - [1, 0]
  (comment [0, 0] - [0, 11]))
```

PR #251, opened 2023-05-27
Nested block comments

susliko:
Fixes https://github.com/tree-sitter/tree-sitter-scala/issues/226

Problem
-------
Nested block comments are not parsed causing an error, e.g.
```
/* /* */ */
```

Solution
-------
Leave `$.comment` for single-line comments.
Introduce `$.block_comment` for block-comments.
Make `$.block_comment` out of several tokens to be able to detect nested block comments

In the future we'll probably want to add support for scala-cli directives inside `$.comment`and scaladoc annotations inside `$.block_comment`  to provide syntax highlighting


grammar.js
```diff
@@ -1309,8 +1310,8 @@ module.exports = grammar({
       seq(
         // opchar
         /[\-!#%&*+\/\\:<=>?@\u005e\u007c~\p{Sm}\p{So}]/,
-        // opchar minus slash
-        /[\-!#%&*+\\:<=>?@\u005e\u007c~\p{Sm}\p{So}]/,
+        // opchar minus slash and asterisk
+        /[\-!#%&+\\:<=>?@\u005e\u007c~\p{Sm}\p{So}]/,
```
eed3si9n, line 1314: Would this cause issue matching on `**`?

susliko: Somehow it doesn't: https://github.com/tree-sitter/tree-sitter-scala/pull/251/commits/8f1f347918b6acce1b0718d79c26f2ae9dfb92b9

susliko: But this change is essential for this PR
Without it the comment start in the following code is parsed as operator_identifier
```scala
object X:
  /** foo */
```

eed3si9n: I guess `**` would match `repeat(/[\-!#%&*+\/\\:<=>?@\u005e\u007c~\p{Sm}\p{So}]/),` of the `choice` branch.

eed3si9n (approved):
Thanks!

Issue #253, opened 2023-05-27
Pattern declaration/definition of val/var


eed3si9n:
### Commit of tree-sitter-scala you tested this on

5a6fcbdddd015169f148d916e246e8faf90cc940

### A code sample showing the error

```scala
val res1, res2 = ArrayBuilder.make[A]
```

### Show the error node

```
scala/collection/ArrayOps.scala	8 ms	(ERROR [1486, 9] - [1486, 14])
```

### Where are you experiencing this error?

smoke test

susliko:
Didn't even know it was a valid Scala syntax, lol

eed3si9n:
Here's my PR for this -https://github.com/tree-sitter/tree-sitter-scala/pull/254

PR #254, opened 2023-05-28
val defn with ids

eed3si9n:
Fixes https://github.com/tree-sitter/tree-sitter-scala/issues/253

Problem
-------
`val` definitions with multiple identifiers are not supported.

Solution
--------
This implements `identifiers`, which acts like a pattern.


ckipp01 (approved):
Looks good, LGTM!

Issue #265, opened 2023-05-28
Refined types


susliko:
### Commit of tree-sitter-scala you tested this on

374709c

### A code sample showing the error

```scala
class A
object A:
  val a: A {
    val b: Int
  } = ???
```

### Show the error node

```
(compilation_unit [0, 0] - [5, 0]
  (class_definition [0, 0] - [0, 7]
    name: (identifier [0, 6] - [0, 7]))
  (object_definition [1, 0] - [5, 0]
    name: (identifier [1, 7] - [1, 8])
    body: (template_body [1, 8] - [5, 0]
      (val_definition [2, 2] - [4, 9]
        pattern: (identifier [2, 6] - [2, 7])
        (ERROR [2, 9] - [2, 10])
        type: (structural_type [2, 11] - [4, 3]
          (val_declaration [3, 4] - [3, 14]
            name: (identifier [3, 8] - [3, 9])
            type: (type_identifier [3, 11] - [3, 14])))
        value: (operator_identifier [4, 6] - [4, 9])))))
```

### Where are you experiencing this error?

Lots of code in scala compiler, [ClassfileParser.scala](https://github.com/scala/scala/blob/4905801fc161300ba20bae4f7adf045b1698e831/src/compiler/scala/tools/nsc/symtab/classfile/ClassfileParser.scala#L47) for instance

eed3si9n:
I noticed that we're missing this as well. Here's my PR for this - https://github.com/tree-sitter/tree-sitter-scala/pull/266

PR #266, opened 2023-05-29
Refinement

eed3si9n:
Fixes https://github.com/tree-sitter/tree-sitter-scala/issues/265

Problem
-------
Compound type with refinement (structural type with a parent) doesn't parse.

Solution
--------
This adds support for it.
It collides with context bounds `A: B: C`, so the precedence is set to -1.


script/smoke_test.sh
```diff
@@ -3,9 +3,9 @@
 # This is an integration test to generally check the quality of parsing.
 
 SCALA_SCALA_LIBRARY_EXPECTED=95
-SCALA_SCALA_COMPILER_EXPECTED=84
+SCALA_SCALA_COMPILER_EXPECTED=87
```
eed3si9n, line 6: Scala 2 compiler bumps to `87%` with this change.

script/smoke_test.sh
```diff
@@ -3,9 +3,9 @@
...
 DOTTY_COMPILER_EXPECTED=81
-SYNTAX_COMPLEXITY_CEILING=2300
+SYNTAX_COMPLEXITY_CEILING=2800
```
ckipp01, line 8: Do we have an idea of how much is too much here? Like when should it cause refactoring to account for the bump?

ckipp01 (approved):
Looks good to me! LGTM!

script/smoke_test.sh
eed3si9n: I guess when we observe compilation time or memory usage to degrade?

PR #269, opened 2023-05-29
Refactor `$.compilation_unit`, optimize grammar

susliko:
Summary
-------
`$.compilation_unit` is now a sequence of top-level statements separated by `$.semicolon`

This change has two effects:
- Grammar optimization:
  - ~10% faster generation time
  - lower number of parser states ([before](https://gist.github.com/susliko/f950b997a98c54bbfd88969a949346fd), [after](https://gist.github.com/susliko/236a85dce46219c5868c494d7f5cf629))
  - parser size reduction from 43M to 36M
- It seems to me, that handling `$._automatic_semicolon` on the top level is a prerequisite to support top-level expressions (https://github.com/tree-sitter/tree-sitter-scala/issues/198) and leading infix operators (https://github.com/tree-sitter/tree-sitter-scala/issues/141)


susliko:
I've tested the hypothesis: this change makes it straightforward to allow top-level expressions (while still passing all test suites)

script/smoke_test.sh
```diff
@@ -3,9 +3,9 @@
 # This is an integration test to generally check the quality of parsing.
 
 SCALA_SCALA_LIBRARY_EXPECTED=95
-SCALA_SCALA_COMPILER_EXPECTED=87
+SCALA_SCALA_COMPILER_EXPECTED=86
```
susliko, line 6: There's a degradation in a single file: 

[StdAttachments.scala#L214](https://github.com/scala/scala/blob/4905801fc161300ba20bae4f7adf045b1698e831/src/compiler/scala/tools/nsc/typechecker/StdAttachments.scala#L214)                 	5 ms	(ERROR [213, 13] - [214, 5]), which is minimized to 
```scala
package x

trait A {
  self: A =>
  import a._
}

trait B {
  self: B =>

  class A() 
  val x = 1
}
```
It seams feasible to me to  investigate the issue individually after this PR gets merged

susliko:
PR is ready for review!

cc @eed3si9n @ckipp01

eed3si9n (approved):
Good to see reduction in complexity. Looking forward to the further changes enabled by this.

ckipp01 (approved):
This looks great @susliko, thanks!

PR #271, opened 2023-05-30
Top-level expressions

susliko:
Resolves https://github.com/tree-sitter/tree-sitter-scala/issues/198

Problem
-------
Top-level expressions are not supported. However, it's a valid case for Scala scripts or sbt configurations

Solution
-------
Now that AUTOMATIC_SEMICOLON tokens are handled at the top-level enabling top-level expressions is done by including `$.expression` into `$._top_level_definition`


eed3si9n (approved):
nice work!

Issue #189, opened 2023-02-07
Feature request: support user-directives.


rochala:
With Scala-CLI probably becoming the default Scala driver, I think adding support for using directives is an important feature to have. The main benefit would most likely be highlighting.

ckipp01:
Thanks @rochala! I have actually been thinking the same thing, especially since I've seen that the VS Code syntax supports it as well.

susliko:
Was implemented in https://github.com/tree-sitter/tree-sitter-scala/pull/273

PR #273, opened 2023-05-31
Support Scala CLI using directives

susliko:
Implements https://github.com/tree-sitter/tree-sitter-scala/issues/189

Summary
-------
Single line `$.comment` is split into a choice of `$._comment_text` and `$.using_directive` to parse a structure of comments like
```
//> using scala 3.3.0
```

Highlighted example:

image
with highlight queries:
```
(using_directive_key) @parameter
(using_directive_value) @string
```

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -1,4 +1,6 @@
 const PREC = {
+  comment: 0,
+  using_directive: 1,
   control: 1,
   stable_type_id: 2,
   binding_decl: 2,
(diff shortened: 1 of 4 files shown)


grammar.js
```diff
@@ -1580,7 +1582,21 @@ module.exports = grammar({
...
 
-    comment: $ => token(seq('//', /.*/)),
+    comment: $ => seq(token('//'), choice(
+      $.using_directive,
+      $._comment_text,
+    )),
+
+    _comment_text: $ => token(prec(PREC.comment, /.*/)),
+
+    using_directive: $ => seq(
```
eed3si9n, line 1592: Could you also add a test for it too here plz? - https://github.com/tree-sitter/tree-sitter-scala/blob/master/test/highlight/basics.scala

susliko: Added to a separate file `comments.scala`

eed3si9n (approved):
Thanks @susliko!

Issue #54, opened 2022-06-16
Allow block before lambda_expression


ptwales:
This is sort of not conventional but is valid:
```scala
val foo: Int => Int = { x => 
  val y = 2 * x
  y * y
}
```

Output from playground:

```
val_definition [0, 0] - [3, 1]
  pattern: identifier [0, 4] - [0, 7]
  type: function_type [0, 9] - [0, 19]
    parameter_types: parameter_types [0, 9] - [0, 12]
      type_identifier [0, 9] - [0, 12]
    return_type: type_identifier [0, 16] - [0, 19]
  value: block [0, 22] - [3, 1]
    lambda_expression [0, 24] - [1, 15]
      identifier [0, 24] - [0, 25]
      assignment_expression [1, 2] - [1, 15]
        left: postfix_expression [1, 2] - [1, 7]
          identifier [1, 2] - [1, 5]
          identifier [1, 6] - [1, 7]
        right: infix_expression [1, 10] - [1, 15]
          left: integer_literal [1, 10] - [1, 11]
          operator: operator_identifier [1, 12] - [1, 13]
          right: identifier [1, 14] - [1, 15]
    infix_expression [2, 2] - [2, 7]
      left: identifier [2, 2] - [2, 3]
      operator: operator_identifier [2, 4] - [2, 5]
      right: identifier [2, 6] - [2, 7]
```
I don't know what this should parse to but `val y` should be a `val_definition` and not a `post_fix_expression`

ptwales:
for the record this is also valid and the same:
```scala
val foo: Int => Int = { 
  x => val y = 2 * x
  y * y
}
```

susliko:
Resolved by https://github.com/tree-sitter/tree-sitter-scala/pull/277!

PR #277, opened 2023-06-05
The Gordian Knot

susliko:
## A startling discovery
changing `$.lambda_expression` body from `$._block` to `$._indentable_expression` results in
* 11-fold decrease in parser generation times (from 110s to 10s on my laptop)
* parser size reduction from 35M to 22M
* resolution of https://github.com/tree-sitter/tree-sitter-scala/issues/54

However, it brings the following problems:
### Problem1: conflict of `$.lambda_expression` and `$.self_type`
Self-type in
```scala
trait A {
  self =>
  val a = 1
}
```
is parsed as lambda expression. Seems to be not much of a problem, as it can be resolved by disabling `$.lambda_expression` in `$.templabe_body` (that's also how it's  in the ebnf)

### Problem2: multi-line lambdas  without newline before closing paren
 ```scala
x.map((x) => 
 if a then b.c)
```

fails to be parsed, because `$._indent` token is emmitted (because it's valid just before the start of `if`, but `$._outdent` is not.
 ```scala
x.map((x) => 
 if a then b.c
)
```
is, on contrary, parsed correctly because of the newline.
For this issue - I can't come up to the solution and will be glad of any ideas


eed3si9n:
```scala
scala> trait A
// defined trait A

scala> foo(new A:
     |   def x = 1)
val res0: A = anon$1@50df37eb
```

so at least in this case, the closing parenthesis works as an outdent, so maybe we should support that in a situation where the parser is looking for an outdent.

eed3si9n:
Here's my PR for the outdent - https://github.com/tree-sitter/tree-sitter-scala/pull/279

SethTisue:
😮

susliko:
@eed3si9n Thank you for the fix of outdents! This helped to solve remaining issues in this PR, it's ready for review 
Key results from the initial description remained:
* ~11-fold decrease in parser generation times with 1Gb max memory usage ceiling
* parser size reduction from 41M to 22M with 1300 as grammar complexity ceiling

.github/workflows/ci.yml
```diff
@@ -66,30 +66,13 @@ jobs:
...
       - name: Generate parser from scratch and test it
-        if: ${{ runner.os == 'Linux' }}
+        if: ${{ runner.os == 'Linux' || needs.changedfiles.outputs.c }}
         shell: bash
         run: |
           npm install
           npm run build
           npm test
 
-      - name: Check parser
```
susliko, line 76: This change would be removed after https://github.com/tree-sitter/tree-sitter-scala/pull/287 resolves

susliko, line 76: This change will be reverted after https://github.com/tree-sitter/tree-sitter-scala/pull/287 resolves

eed3si9n: Here's my attempt at this - https://github.com/tree-sitter/tree-sitter-scala/pull/288

kanazr666 (approved):

.github/workflows/ci.yml
susliko: Rebased on master, reverted this change

eed3si9n (approved):
Thanks @susliko!

ckipp01:
Really great job on this @susliko!

PR #278, opened 2023-06-06
Use prettier to format JavaScript file

eed3si9n:
Problem
--------
There are some whitespaces at the end of the line kind of things.

Solution
--------
This implements formatting using [Prettier](https://prettier.io/).

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -21,22 +21,14 @@ const PREC = {
   field: 8,
   macro: 10,
   binding: 10,
-}
+};
 
 module.exports = grammar({
-  name: 'scala',
+  name: "scala",
 
-  extras: $ => [
-    /\s/,
-    $.comment,
-    $.block_comment,
-  ],
+  extras: $ => [/\s/, $.comment, $.block_comment],
 
-  supertypes: $ => [
-    $.expression,
-    $._definition,
-    $._pattern,
-  ],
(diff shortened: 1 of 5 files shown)


grammar.js
ckipp01, line 1: Since this is a pretty huge diff here, can we add one more commit adding in a [`.git-blame-ignore-revs`](https://moxio.com/blog/ignoring-bulk-change-commits-with-git-blame/) that ignores the reformat commit?

ckipp01 (approved):
Looks good to me!

grammar.js
```diff
@@ -1341,292 +1341,280 @@ module.exports = grammar({
...
+      seq(
+        token.immediate(prec(PREC.using_directive, ">")),
+        token("using"),
+        $.using_directive_key,
+        $.using_directive_value,
+      ),
     using_directive_key: $ => token(/[^\s]+/),
     using_directive_value: $ => token(/.*/),
 
-    block_comment: $ => seq(
```
susliko, line 1601: In such cases, multi-line formatting is intentional and (very subjectively) more easily comprehended. I'd suggest disable auto-wrapping by setting `printWidth` to null

eed3si9n: Yea. Unfortunately, I don't think there's a way to preserve end-of-line.
It's either going to make all the lines long, or always add EOL.

susliko: Ah, okay

PR #279, opened 2023-06-06
Allow closing paren/bracket to outdent

eed3si9n:
Problem
-------
Braceless syntax can terminate with closing paren or bracket of the outer context, but we don't support that.

Solution
--------
This changes the scanner so when an outdent is expected, closing paren or bracket can inject an outdent.


ckipp01 (approved):
Huh, I never even thought of this.

eed3si9n:
Yea, it's a weird corner case of parsing specific to Scala 3 syntax.

Issue #262, opened 2023-05-28
Repeated params in matches


susliko:
### Commit of tree-sitter-scala you tested this on

374709c

### A code sample showing the error

```scala
object A:
  x match {
    case Array(a: Type, _@_*) => y
  }
```

### Show the error node

```
(compilation_unit [0, 0] - [6, 0]
  (object_definition [0, 0] - [6, 0]
    name: (identifier [0, 7] - [0, 8])
    body: (template_body [0, 8] - [6, 0]
      (match_expression [1, 2] - [3, 3]
        value: (identifier [1, 2] - [1, 3])
        body: (case_block [1, 10] - [3, 3]
          (case_clause [2, 4] - [3, 2]
            pattern: (case_class_pattern [2, 9] - [2, 29]
              type: (type_identifier [2, 9] - [2, 14])
              pattern: (typed_pattern [2, 15] - [2, 22]
                pattern: (identifier [2, 15] - [2, 16])
                type: (type_identifier [2, 18] - [2, 22]))
              pattern: (infix_pattern [2, 24] - [2, 27]
                left: (wildcard [2, 24] - [2, 25])
                operator: (operator_identifier [2, 25] - [2, 26])
                right: (wildcard [2, 26] - [2, 27]))
              (ERROR [2, 27] - [2, 28]
                (operator_identifier [2, 27] - [2, 28])))
            body: (identifier [2, 33] - [2, 34])))))))
```

### Where are you experiencing this error?

[BackendUtils.scala](https://github.com/scala/scala/blob/4905801fc161300ba20bae4f7adf045b1698e831/src/compiler/scala/tools/nsc/backend/jvm/analysis/BackendUtils.scala#L562)

eed3si9n:
Here's my PR for this - https://github.com/tree-sitter/tree-sitter-scala/pull/281

PR #281, opened 2023-06-06
Repeat pattern

eed3si9n:
Fixes https://github.com/tree-sitter/tree-sitter-scala/issues/262

Problem
-------
1. Repeat pattern with `*` doesn't parse.
2. Capture pattern `@` doesn't capture wildcard.

Solution
--------
1. This implements support for `*` in pattern.
2. This also allows using `_@`.


susliko:
Looks good to me!

PR #282, opened 2023-06-06
Added block comments to the comment highlighting query

domaspoliakas:
As per a quick conversation on nvim-metals discord channel - block comments didn't get highlighting query added to them when they were split into their own thing, as per @keynmol :

> I also found the commit where it slipped through the cracks - the comment node got split but the highlighting query wasn't added: https://github.com/tree-sitter/tree-sitter-scala/commit/693c651b509e2654c237a68e52080f0211b4f5bf

This PR rectifies that and adds tests.

Let me know if there's anything else needed here!

### DIFF
--- a/test/highlight/comments.scala
+++ b/test/highlight/comments.scala
@@ -2,3 +2,11 @@
 //  ^keyword
 //        ^parameter
 //              ^string
+
+/*
+ * Beep boop
+ */
+// <- comment
+
+// Single line comment
+// <- comment
--- a/queries/scala/highlights.scm
+++ b/queries/scala/highlights.scm
@@ -236,6 +236,7 @@
 "return" @keyword.return
 
 (comment) @comment @spell
+(block_comment) @comment @spell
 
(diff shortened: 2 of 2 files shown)


test/highlight/comments.scala
```diff
@@ -2,3 +2,11 @@
 //  ^keyword
 //        ^parameter
 //              ^string
+
+/*
+ * Beep boop
+ */
+// <- comment
```
keynmol, line 9: Nice, was just about to ask but then [read the docs](https://tree-sitter.github.io/tree-sitter/syntax-highlighting#unit-testing):

> Arrow: <- this will test the following selector against the scope on the most recent non-test line. It will test it at the same column as the comment character is in.

susliko:
I think, these queries should also be added to [nvim-treesitter](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/queries/scala/highlights.scm)

ckipp01 (approved):
Thanks @domaspoliakas!

Issue #258, opened 2023-05-28
Structural types in `extends` clause


susliko:
### Commit of tree-sitter-scala you tested this on

374709c

### A code sample showing the error

```scala
object namerFactory extends {
  val a: A = b
} 
```

### Show the error node

```
(compilation_unit [0, 0] - [4, 0]
  (object_definition [0, 0] - [2, 1]
    name: (identifier [0, 7] - [0, 19])
    (ERROR [0, 20] - [0, 27])
    body: (template_body [0, 28] - [2, 1]
      (val_definition [1, 2] - [1, 14]
        pattern: (identifier [1, 6] - [1, 7])
        type: (type_identifier [1, 9] - [1, 10])
        value: (identifier [1, 13] - [1, 14])))))
```

### Where are you experiencing this error?

Reading Scala compiler code is like reading some ancient scrolls:
[Analyzer.scala](https://github.com/scala/scala/blob/4905801fc161300ba20bae4f7adf045b1698e831/src/compiler/scala/tools/nsc/typechecker/Analyzer.scala#L38)

PR #285, opened 2023-06-07
Fix structural type in extends

eed3si9n:
Fixes https://github.com/tree-sitter/tree-sitter-scala/issues/258

Problem
-------
Structural type in extends clause fail to parse.

Solution
--------
This adds structural_type to the simple_type.


ckipp01 (approved):
👍🏼

PR #286, opened 2023-06-07
fix: make sure queries are copied to the right place

ckipp01:
I noticed that the queries in nvim-treesitter and this repo have
diverged, but we never got an error here. I think this was a regression
introduced in
https://github.com/tree-sitter/tree-sitter-scala/commit/7d348f51e442563f4ab2b6c3e136dac658649f93.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,7 +98,7 @@ jobs:
       - name: copy nvim-treesitter queries
         if: ${{ runner.os == 'Linux' }}
         shell: bash
-        run: cp ./nvim_treesitter/queries/scala/*.scm ./queries/
+        run: cp ./nvim_treesitter/queries/scala/*.scm ./queries/scala/
 
       - name: Check if queries are out of sync with nvim-treesitter
         if: ${{ runner.os == 'Linux' }}
--- a/queries/scala/highlights.scm
+++ b/queries/scala/highlights.scm
@@ -235,8 +235,13 @@
 
 "return" @keyword.return
 
-(comment) @comment @spell
-(block_comment) @comment @spell
+[
+  (comment)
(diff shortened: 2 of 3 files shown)


eed3si9n (approved):
Thanks!

ckipp01:
So I sort of forgot about this, but it's failing right now and it's _somewhat_ expected. Basically some of the improvements that were added by @ghostbuster91  in https://github.com/tree-sitter/tree-sitter-scala/pull/196 are great and they follow the [upstream locals](https://tree-sitter.github.io/tree-sitter/syntax-highlighting#local-variables) way of doing things. However, this is one area where nvim-treesitter [differs](https://github.com/nvim-treesitter/nvim-treesitter/pull/4567#issuecomment-1484910842). So if we sync the local queries here, it won't be quite as accurate. I'm sort of inclined to update the check to only check against the actual `hightlights.scm` file since the `locals.scm` will be different. Anyone have thoughts on this?

susliko:
I'd suggest being explicit about the fact that queries are system-specific.  For example, we could have a structure like
```
queries
  (highlights|locals).scm - vanilla tree-sitter
integrations
  nvim
    queries
      (highlights|locals|etc.).scm
```

And check only queries/* in the Ci

ghostbuster91:
> I'd suggest being explicit about the fact that queries are system-specific

:+1: 

I would not revert the changes in the `locals.scm` as there might be other editors that use treestitter (e.g. kakune, helix), that could benefit from these definitions.

ckipp01:
Just commenting that I haven't forgotten about this, 😬 just been busy lately with other things. I'll try to return soon.

ckipp01:
Closing this just because I won't have time to work on it.

PR #287, opened 2023-06-07
Failing CI on MacOS/Windows?

susliko:
Haven't found a button to run the workflow manually, hence this PR
I suspect that CI is failing for MacOS and Windows

### DIFF
--- a/src/scanner.c
+++ b/src/scanner.c
@@ -4,6 +4,7 @@
 #include <string.h>
 #include <wctype.h>
 
+// Just a comment to trigger CI on MacOS and Windows
 enum TokenType {
   AUTOMATIC_SEMICOLON,
   INDENT,


susliko:
Indeed, it fails
I feel like test jobs for MacOS and Windows operate on some old version of grammar. Particulary, they fail to parse structural types in extend clauses brought by https://github.com/tree-sitter/tree-sitter-scala/pull/285

cc @eed3si9n @ckipp01

susliko:
Gotcha: the phase that generates parser out of the grammar is only run on Linux, and Windows/MacOS use the `parser.c` from the repo, which is stale 

I think we should generate parser out of the grammar every time we run tests

susliko:
> If I recall correctly, at some point @keynmol, @ckipp01, and I decided that running checks on Linux is good enough, except for the external scanner C code, which we run test on all 3 OSes.

I think, running MacOS/Windows tests only when external scanner is changed is okay as long as we run them using freshly generated parser

eed3si9n:
> as long as we run them using freshly generated parser

+1

susliko:
Could you give me a hint on the purpose of the [Check parser](https://github.com/susliko/tree-sitter-scala-1/blob/7da300e016d0d269eab3cba886e886f28d295749/.github/workflows/ci.yml#L76) job? 🤔

eed3si9n:
> Could you give me a hint on the purpose of the [Check parser](https://github.com/susliko/tree-sitter-scala-1/blob/7da300e016d0d269eab3cba886e886f28d295749/.github/workflows/ci.yml#L76) job? 🤔

To prevent someone committing a malicious C code, we test that _if_ C sources are checked in there should be no `git diff` output after running generate.

susliko:
Is there a reason why it's disabled for Linux environment?

susliko:
> When I first introduced it on https://github.com/tree-sitter/tree-sitter-scala/commit/95c14376350613920eccfe5ec37ceb1c582721b0 it was Linux only, then it became non-Windows on https://github.com/tree-sitter/tree-sitter-scala/commit/fc77eb0c5d1e0159bdb22296d5221b2b624bd024, then should've been back to Linux only at https://github.com/tree-sitter/tree-sitter-scala/commit/8aface51ab1ab70bcd23617d4edf50a6b0f5f103, but there was a typo on == vs !=, I think.

The check fails for this PR, not sure why

eed3si9n:
Closing this in favor of https://github.com/tree-sitter/tree-sitter-scala/pull/288

PR #288, opened 2023-06-08
Fix C code fidelity check

eed3si9n:
See https://github.com/tree-sitter/tree-sitter-scala/pull/287

Problem
-------
There's C code fidelity check in the CI process, but over time it seems to have broken.
`needs.changedfiles.outputs.c` gets triggered for either `src/scanner.c` or the generated `src/parser.c`, then the fidelity check logic expects that `git diff` will be empty.
This actually won't work if you changed both `src/parser.c` and `grammar.js` at the same time.

Solution
--------
The fix is to make a new `changedfiles.outputs.gen`. Now the new
behavior is:
1. If any C code were changed (including `src/parser.c`), run full tests on all 3 Oses.
2. If non-`src/parser.c` C code were changed, run fidelity check on Linux.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,7 @@ jobs:
     outputs:
       all: ${{ steps.changes.outputs.all}}
       c: ${{ steps.changes.outputs.c }}
+      gen: ${{ steps.changes.outputs.gen }}
     steps:
       - name: checkout tree-sitter-scala
         uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -20,7 +20,8 @@ jobs:
         id: changes
         run: |
           echo "all=$(git diff --name-only --diff-filter=ACMRT ${{ github.event.pull_request.base.sha }} ${{ github.sha }} | xargs)" >> $GITHUB_OUTPUT
-          echo "c=$(git diff --name-only --diff-filter=ACMRT ${{ github.event.pull_request.base.sha }} ${{ github.sha }} | grep '\.\(c\|h\)$' | xargs)" >> $GITHUB_OUTPUT
+          # exclude src/scanner.c from the changed C code, because we'd normally review it.
+          echo "c=$(git diff --name-only --diff-filter=ACMRT ${{ github.event.pull_request.base.sha }} ${{ github.sha }} | grep '\.\(c\|h\)$' | grep -v 'src/scanner.c' | xargs)" >> $GITHUB_OUTPUT
```
susliko, line 24: If we stop tracking changes at `src/scanner.c`, `needs.changedfiles.outputs.c` won't be set and tests in non-Linux environments won't be triggered. Is that a desired behavior?

eed3si9n: Gaa you're right. I guess I can make another category like `gen`

eed3si9n: Addressed in https://github.com/tree-sitter/tree-sitter-scala/pull/288/commits/5029432a9ec40ebc88b8aee2ef097678c437df06

susliko:
Great! 🎉

susliko (approved):

Issue #257, opened 2023-05-28
Definitions with multiple param lists on different lines


susliko:
### Commit of tree-sitter-scala you tested this on

374709c

### A code sample showing the error

```scala
object A:
  def a(a: Int)
       (b: Int => Int): String = ???
```

Similar behavior for class definitions

### Show the error node

```
(compilation_unit [0, 0] - [3, 0]
  (object_definition [0, 0] - [3, 0]
    name: (identifier [0, 7] - [0, 8])
    body: (template_body [0, 8] - [3, 0]
      (function_declaration [1, 2] - [1, 15]
        name: (identifier [1, 6] - [1, 7])
        parameters: (parameters [1, 7] - [1, 15]
          (parameter [1, 8] - [1, 14]
            name: (identifier [1, 8] - [1, 9])
            type: (type_identifier [1, 11] - [1, 14]))))
      (ERROR [2, 7] - [2, 23]
        (bindings [2, 7] - [2, 22]
          (binding [2, 8] - [2, 21]
            name: (identifier [2, 8] - [2, 9])
            type: (function_type [2, 11] - [2, 21]
              parameter_types: (parameter_types [2, 11] - [2, 14]
                (type_identifier [2, 11] - [2, 14]))
              return_type: (type_identifier [2, 18] - [2, 21])))))
      (assignment_expression [2, 24] - [2, 36]
        left: (identifier [2, 24] - [2, 30])
        right: (operator_identifier [2, 33] - [2, 36])))))
```

PR #295, opened 2023-06-09
Multi-line parameter lists

susliko:
Fixes https://github.com/tree-sitter/tree-sitter-scala/issues/257

Problem
-------
Parameter lists spreading multiple lines are not handled:
```scala
class A
():
  def b
    (): Int = 1
```

Solution
-------
Match `$._automatic_semicolon` in `$._function_constructor` and `$.class_parameters`

Side-change: reuse `$._function_declaration` in `$.function_definition`


eed3si9n (approved):
Nice. Thanks @susliko!

susliko:
@eed3si9n wow, so fast! 🚀

PR #297, opened 2023-06-10
Structural types in declarations

susliko:
Problem
-------
Structural types in declarations are not parsed:
```scala
val x: F { val y: Int }
```

Solution
-------
Remove -1 precedence for `$._structural_type` branch of  `$.compount_type`, resolve conflict with `$._type` explicitly

Extra changes
-------
Removed some unused conflict resolutions


corpus/types.txt
```diff
@@ -198,8 +198,10 @@ type G = H[A1] {
   def foo[A2 <: Stepper[_]]: A2
 }
 
-type I = F1[A1:
-  def foo: A2]
+type I = F1{
+  def foo: A}
```
eed3si9n, line 202: I added this test to demonstrate the use of braceless syntax, in a type parameter position, outdented by a closing bracket. Not being able to parse it seems like a regression, assuming that what I wrote is in fact a legal Scala.

eed3si9n (changes requested):

corpus/types.txt
susliko: Hm, could you please proved a compiling Scala example with such code?
e.g.
```scala
object X:
  trait F1[T]
  type I = F1[A1:
    def foo: A2]
```
would fail to compile

eed3si9n:
```scala
scala> type I = List[AnyRef:
     |   def toString: String]
// defined alias type I = List[AnyRef{def toString: String}]
```

Here's I guess a `List` type whose element must be a structural type that implements `toString`

susliko: Thanks!
I had scalac 3.1.2, which was old enough not to compile this

Reverted this change

eed3si9n (approved):
Thanks!

PR #301, opened 2023-06-10
Issue OUTDENT with extra spaces before brackets

susliko:
Problem
-------
```scala
{ x =>
  if (a) b.c }
```
is not parsed because OUTDENT token is not emitted in a presense of space characters before the closing bracket

Solution
-------
Reorder space-handling and outdent-ing logic in external scanner


script/smoke_test.sh
```diff
@@ -2,7 +2,7 @@
 
 # This is an integration test to generally check the quality of parsing.
 
-SCALA_SCALA_LIBRARY_EXPECTED=99
+SCALA_SCALA_LIBRARY_EXPECTED=100
```
eed3si9n, line 5: 💯🎉

src/scanner.c
```diff
@@ -109,6 +109,16 @@ bool tree_sitter_scala_external_scanner_scan(void *payload, TSLexer *lexer,
   int indentation_size = 0;
   LOG("scanner was called at column: %d\n", lexer->get_column(lexer));
 
+  while (iswspace(lexer->lookahead)) {
```
eed3si9n, line 112: LGTM. Thanks!

eed3si9n (approved):

Issue #305, opened 2023-06-12
Nested comments break syntax hightlight in nvim treesitter


antosha417:
### Commit of tree-sitter-scala you tested this on

latest

### A code sample showing the error

```scala
package bla

// /*
//  * This is awesome comment
//  */
// case class Foo()

case class Bla()
```
<img width="738" alt="Screenshot 2023-06-12 at 13 03 07" src="https://github.com/tree-sitter/tree-sitter-scala/assets/14187674/a6a66053-b634-486a-9f39-38ad16dc202b">
<img width="682" alt="Screenshot 2023-06-12 at 13 03 25" src="https://github.com/tree-sitter/tree-sitter-scala/assets/14187674/48be4230-2bbe-4f61-83d2-ed826e5e0758">

### Show the error node

```
package_clause [0, 0] - [0, 11]
  name: package_identifier [0, 8] - [0, 11]
    identifier [0, 8] - [0, 11]
ERROR [2, 0] - [9, 0]
  comment [3, 0] - [3, 29]
  comment [4, 0] - [4, 6]
  comment [5, 0] - [5, 19]
```

### What do you expect the tree to look like

I want syntax highlighting to work even when there are nested comments

### Where are you experiencing this error?

nvim-treesitter

susliko:
@antosha417 Thanks for reporting!

PR #306, opened 2023-06-12
Better comments

susliko:
Resolves https://github.com/tree-sitter/tree-sitter-scala/issues/305

Problem
-------
The following comment combinations are not parsed:
1. ``` // /*
// *
// */
```
2. ```
/* // */
```
Solution
-------
- set higher lexing priority for `$._comment_text` token
- include "//" as an alternative to the contents of `$.block_comment`


eed3si9n (approved):
Thanks @susliko!

Issue #310, opened 2023-06-17
ERROR on parsing `HkTypeParamClause ‘=>>’ Type`.


KaranAhlawat:
### Commit of tree-sitter-scala you tested this on

a5c4c29

### A code sample showing the error

```scala
class X extends Y[
  [A <: B] =>> C,
] {}
```

### Show the error node

```
(class_definition class body: (identifier)
 (extends_clause extends
  arguments: 
   (generic_type type: (type_identifier)
    type_arguments: 
     (type_arguments [
      (ERROR [)
      (infix_type left: (type_identifier) operator: (operator_identifier)
                  right: (type_identifier))
      (ERROR ] = > >)
      ])))
 (template_body { }))
```

### Where are you experiencing this error?

treesit (GNU Emacs 29+)

susliko:
@KaranAhlawat Thanks for reporting!

KaranAhlawat:
Fixed in #312. Closing.

PR #311, opened 2023-06-20
Add support for Type Lambdas

KaranAhlawat:
First PR, please bear with me.
This is in relation to #310. I later came to learn that apparently these are called type lambdas.

The implementation of the rule itself is trivial, looking at the official Scala 3 syntax grammar. Hence, I have added the necessary changes.

My only concerns are
1. I think I've covered all the cases, but since I'm new to this, I'm not a 100% sure
2. I'm not sure how to run the test against the official compiler code-bases that needs to be updated.

Appreciate any help and guidance, thank you!

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -344,6 +344,7 @@ module.exports = grammar({
           $.covariant_type_parameter,
           $.contravariant_type_parameter,
           $._type_parameter, // invariant type parameter
+          $.type_lambda,
         ),
       ),
 
@@ -788,6 +789,7 @@ module.exports = grammar({
         $._annotated_type,
         $.literal_type,
         $._structural_type,
+        $.type_lambda,
       ),
 
     _annotated_type: $ => prec.right(choice($.annotated_type, $._simple_type)),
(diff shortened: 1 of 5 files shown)


src/parser.c
eed3si9n, line 1: Thanks for the contribution. Could you exclude changes to `src/` from the PR please? We have a GitHub bot that will add these automatically to prevent someone adding malicious C code.

grammar.js
```diff
@@ -929,6 +931,15 @@ module.exports = grammar({
...
     _type_identifier: $ => alias($._identifier, $.type_identifier),
 
+    type_lambda: $ =>
+      seq(
+        "[",
+        trailingCommaSep1($._type_parameter),
+        "]",
+        "=>>",
+        field("return_type", $._type),
+      ),
```
eed3si9n, line 941: This LGTM

src/parser.c
KaranAhlawat: I'll try to figure out how to do that. Not entirely sure, but will update when it's done.

PR #321, opened 2023-06-30
Use match? instead of lua-match? for queries

natsukagami:
Improves compatibility with non-`nvim` tree-sitter implementations.
Since we only have very simple patterns, just switching to `match` should be fine.

### DIFF
--- a/queries/scala/highlights.scm
+++ b/queries/scala/highlights.scm
@@ -64,18 +64,18 @@
 ((stable_identifier (identifier) @namespace))
 
 ((import_declaration
-  path: (identifier) @type) (#lua-match? @type "^[A-Z]"))
-((stable_identifier (identifier) @type) (#lua-match? @type "^[A-Z]"))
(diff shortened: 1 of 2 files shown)


susliko:
Hi!
This looks relatable to the discussion in
https://github.com/tree-sitter/tree-sitter-scala/pull/286

natsukagami:
Thanks for pointing me there. I think for these queries they don't really need to be `lua-match`, and using `match` allows both the WAsm and Rust implementations to handle the predicates.

natsukagami:
Failing test doesn't seem right:
```scala
 class C:
 // ^keyword
   fooooo.map: x =>
   // ^type
   //     ^method.call
     x + 1
```
This used to pass due to `lua-match` predicate not being run with `tree-sitter` command. 

I changed the test to
```scala
 class C:
 // ^keyword
 //    ^type
   fooooo.map: x =>
   //     ^method.call
     x + 1
```

susliko (approved):
Looks good to me. Thanks!

PR #324, opened 2023-07-20
fix: handle interpolations without spaces and $$ in interpolated strings

johannescoetzee:
The Scala syntax summary doesn't describe the `$identifier`-type interpolations in interpolated/processed strings correctly. While `a$b_<operator_chars>` is a valid identifier name, in the interpolated string `s"$a$b_<operator_chars>`, this is actually treated as two identifier interpolations `$a` and `$b_`, followed by the string part `<operator_chars>`. 

To fix this, I've added a new rule, `_interpolation_identifier` to describe identifiers in simple interpolations that matches the regex `/[A-Za-z_][A-Z_a-z0-9]/;` instead of the full regex in `_alpha_identifier`. ~This does mean that queries relating to interpolated strings will need to be updated, but the changes to the queries in this repo are minimal~. I've aliased the rule to `$.identifier` instead, so this will work without query changes.

This also includes a smaller change to handle `$$` patterns in interpolated strings (which evaluate to just `$`) without errors. These don't appear in the tree, similar to regular escape sequences.

### DIFF
--- a/corpus/expressions.txt
+++ b/corpus/expressions.txt
@@ -706,9 +706,9 @@ def matchTest(x: Int): String = x match {
             (identifier)
             (interpolated_string
               (interpolation
-                (interpolation_identifier))
+                (identifier))
               (interpolation
-                (interpolation_identifier))))
+                (identifier))))
           (infix_expression
             (identifier)
             (operator_identifier)
--- a/corpus/expressions.txt
+++ b/corpus/expressions.txt
@@ -706,9 +706,9 @@ def matchTest(x: Int): String = x match {
             (identifier)
             (interpolated_string
               (interpolation
-                (identifier))
+                (dollar_identifier))
               (interpolation
-                (identifier))))
+                (dollar_identifier))))
           (infix_expression
             (identifier)
             (operator_identifier)
--- a/corpus/expressions.txt
+++ b/corpus/expressions.txt
@@ -706,9 +706,9 @@ def matchTest(x: Int): String = x match {
             (identifier)
             (interpolated_string
               (interpolation
-                (dollar_identifier))
+                (interpolation_identifier))
               (interpolation
-                (dollar_identifier))))
+                (interpolation_identifier))))
           (infix_expression
             (identifier)
             (operator_identifier)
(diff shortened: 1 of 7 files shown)


johannescoetzee:
@eed3si9n Done! I think the added test covers everything, but please let me know if I should add more.

eed3si9n (approved):
Thanks @johannescoetzee!

eed3si9n:
**Note** Different editors maintain their of queries for tree-sitter-scala (https://github.com/tree-sitter/tree-sitter-scala/discussions/283), so:
1.  after this lands, we need to propagate the new highlight queries to different repos to take effect
2. however, since this adds a new token `interpolation_identifier`, if the end user somehow gets the updated queries but not the updated tree-sitter-scala, it may crash their editor?

johannescoetzee:
> however, since this adds a new token interpolation_identifier, if the end user somehow gets the updated queries but not the updated tree-sitter-scala, it may crash their editor?

There are 2 potential crashes when updating tree-sitter-scala and the queries: 
1. If the queries are updated first, tree-sitter will crash on the unknown token.
2. If tree-sitter-scala is updated first, then at least nvim-treesitter crashes since `(interpolation (identifier))` is no longer a possible pattern.

eed3si9n:
> If tree-sitter-scala is updated first, then at least nvim-treesitter crashes since (interpolation (identifier)) is no longer a possible pattern.

Could we somehow prevent this by creating a dummy token that can return `(interpolation (identifier))`? By nature of the tree-sitter, the parser becomes available immediately when the `*.c` is checked into the master branch.

johannescoetzee:
It turns out it's possible to alias rules, so I've pushed a change to alias `$._interpolation_identifier` to `$.identifier`. This means queries won't have to be updated at all.

johannescoetzee:
@eed3si9n I realised this morning that this doesn't handle the `$"` dollar escape, so I've updated the `dollar_escape` rule to include that case. To confirm that it's valid, I tested both scala 2 and 3 (results only shown for scala 3)
```
❯ scala
Welcome to Scala 3.3.0 (19.0.2, Java OpenJDK 64-Bit Server VM).
Type in expressions for evaluation. Or try :help.
                                                                                                                                                               
scala> val a = 2
val a: Int = 2
                                                                                                                                                               
scala> s"$"$a"
val res0: String = "2
```

Issue #302, opened 2023-06-10
Scala 2 macros


susliko:
### Commit of tree-sitter-scala you tested this on

3573bf7

### A code sample showing the error

```scala
def a: A =
  macro B.b
```

### Show the error node

```
(compilation_unit [0, 0] - [2, 0]
  (function_definition [0, 0] - [2, 0]
    name: (identifier [0, 4] - [0, 5])
    return_type: (type_identifier [0, 7] - [0, 8])
    body: (indented_block [1, 2] - [2, 0]
      (field_expression [1, 2] - [1, 11]
        value: (identifier [1, 2] - [1, 7])
        (ERROR [1, 8] - [1, 9])
        field: (identifier [1, 10] - [1, 11])))))
```

### What do you expect the tree to look like

Some rules for Scala 2 macro should be introduced

eed3si9n:
here's my PR for this - https://github.com/tree-sitter/tree-sitter-scala/pull/325

PR #325, opened 2023-07-23
Scala 2 macros

eed3si9n:
Fixes https://github.com/tree-sitter/tree-sitter-scala/issues/302

Problem
-------
Scala 2 macro definition is not supported.

Solution
--------
This adds macro_body as an expression.


ckipp01 (approved):
Looks good, thanks @eed3si9n! LGTM

PR #340, opened 2023-08-18
Tidy up scanner and generate w/ master

amaanq:
the only change I might be unsure about is changing the nul byte check to an eof check, I'm assuming scala is sane and doesn't allow for that anywhere but eof

### DIFF
--- a/src/scanner.c
+++ b/src/scanner.c
@@ -25,24 +25,21 @@ void *tree_sitter_scala_external_scanner_create() {
   return createStack();
 }
 
-void tree_sitter_scala_external_scanner_destroy(void *p) {
-  free(p);
+void tree_sitter_scala_external_scanner_destroy(void *payload) {
+  free(payload);
 }
 
-void tree_sitter_scala_external_scanner_reset(void *p) {
-  resetStack(p);
+unsigned tree_sitter_scala_external_scanner_serialize(void *payload, char *buffer) {
+  return serialiseStack(payload, buffer);
 }
 
-unsigned tree_sitter_scala_external_scanner_serialize(void *p, char *buffer) {
-  return serialiseStack(p, buffer);
-}
-
-void tree_sitter_scala_external_scanner_deserialize(void *p, const char *b,
-                                                    unsigned n) {
-  deserialiseStack(p, b, n);
+void tree_sitter_scala_external_scanner_deserialize(void *payload, const char *buffer,
+                                                    unsigned length) {
+  deserialiseStack(payload, buffer, length);
 }
 
 static void advance(TSLexer *lexer) { lexer->advance(lexer, false); }
+
 static void skip(TSLexer *lexer) { lexer->advance(lexer, true); }
 
 static bool scan_string_content(TSLexer *lexer, bool is_multiline, bool has_interpolation) {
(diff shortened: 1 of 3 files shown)


amaanq:
`tree_sitter_LANG_external_scanner_reset` hasn't been a thing for a long time iirc, so I removed that from test-stack as well

ckipp01 (changes requested):
Thanks a lot for looking at all of this @amaanq! It looks like there is a diff for some reason on the checked in code vs what is generated in CI. For a while we were having quite a bit of development happening so we got into the habbit of not checking in any generated code, and instead we have a nightly job that runs which generates everything and then syncs it. So it looks like CI is failing because what you checked in doesn't seem to match what was generated in CI. I'm going to assume that you generated it correctly, but maybe we're not doing something correct in CI, or maybe there is a slight difference in how we're generating?

One solution would be to just remove the generated stuff from the PR and then let the CI take care of it. Also you can see what the diff is [here](https://github.com/tree-sitter/tree-sitter-scala/actions/runs/5898443464/job/15999489975).

amaanq:
I believe it's because you use the node version to generate instead of the rust cli app on master (there's been quite some changes in generation to it lately)

ckipp01:
> I believe it's because you use the node version to generate instead of the rust cli app on master (there's been quite some changes in generation to it lately)

Interesting, I would've expected that generation via Node or the Rust cli would be the same. Is there a preferred way to do it?

amaanq:
Node is a bit lacking because most development happens w/ the Rust CLI, and then node is updated every release, so obviously you miss out on some benefits that master offers

As of now what I remember is

Regex "i" flag support (no more hacks to get this functionality)

Renaming of more sym variables to appropriate names (anon_sym_ for a space becomes anon_sym_SPACE)

Removal of literal special chars in favor of escaped versions, unicode chars now are escaped w/ either \u{4}, \U{8}, or \0 (getting rid of literal null chars in parser.c was the biggest motivator for this)

VT and FF are added to the \s whitespace class, they had minimal impact on parser binary size while parsing code that used it properly (obviously there's some more whitespace chars missing but they killed the binary size/compilation time, so we opted for these 2 for now)

Moving eof assignment in the macro to avoid that annoying unused warning when running npm install

src/scanner.c
```diff
@@ -101,6 +100,16 @@ static bool detect_comment_start(TSLexer *lexer) {
   return false;
 }
 
+static bool scan_word(TSLexer *lexer, const char* const word) {
```
eed3si9n, line 103: Thanks!

eed3si9n:
The first commit looks good to me.

ckipp01:
> You could have the CI use tree-sitter's crate @master, that would always be the most correct and up-to-date

Sure, we can consider that. Thanks for the extra input @amaanq. For now, do you think it's possible to drop the second generation commit and just keep the changes made in the first? Then our CI can pass, we can get this merged in / generated, and then on our end we can discuss switching from npm to rust if it is indeed better.

amaanq:
Sure thing, ts is due for a new release soon anyways

amaanq:
Btw, you guys might be excited to know a few other features upstream CLI has to perhaps incorporate in your CI/testing

- warnings when certain conflicts are unnecessary

- support for an --apply-all-captures flag and --queries-path flag to run queries/highlights/tests w/ the last-wins precedence (like Neovim). Max, I, and others agreed this is the better way, and will be the default in the future

- support for negative assertions by prefixing the assertion with a !

- support for dirs inside corpus (nested testing), previously this wasn't possible, but now in C++ I added C's corpus to ensure *that* passes as well :)

I might've forgotten a few things...

Issue #332, opened 2023-08-08
Anonymous given with `using` clause


susliko:
### Commit of tree-sitter-scala you tested this on

46e08442cd40fb33742594278a5554104cebcb31

### A code sample showing the error

```scala
given (using a: Int) : Int = a
```

### Show the error node

```
(compilation_unit [0, 0] - [1, 0]
  (given_definition [0, 0] - [0, 30]
    return_type: (tuple_type [0, 6] - [0, 20]
      (infix_type [0, 7] - [0, 19]
        left: (infix_type [0, 7] - [0, 15]
          left: (type_identifier [0, 7] - [0, 12])
          operator: (identifier [0, 13] - [0, 14])
          right: (type_identifier [0, 14] - [0, 15]))
        operator: (identifier [0, 16] - [0, 19])
        right: (type_identifier [0, 19] - [0, 19])))
    (ERROR [0, 21] - [0, 26])
    body: (identifier [0, 29] - [0, 30])))
```

### What do you expect the tree to look like

```
(compilation_unit [0, 0] - [1, 0]
  (given_definition [0, 0] - [0, 31]
    parameters: (parameters [0, 7] - [0, 21]
      (parameter [0, 14] - [0, 20]
        name: (identifier [0, 14] - [0, 15])
        type: (type_identifier [0, 17] - [0, 20])))
    return_type: (type_identifier [0, 24] - [0, 27])
    body: (identifier [0, 30] - [0, 31])))
```

eed3si9n:
Here's a PR for this - https://github.com/tree-sitter/tree-sitter-scala/pull/341

PR #341, opened 2023-08-27
Fixes anonymous givens

eed3si9n:
Fixes https://github.com/tree-sitter/tree-sitter-scala/issues/332

Problem
-------
Currently given uses `_function_constructor`, which expects an identifier.
This doesn't work for anonymous given definition that has function-ish things like `using` parameters.

Solution
--------
Create a new `_given_constructor` node with an optional name.


eed3si9n:
Apparently this regresses the smoke test

```
Report written to report-dotty-compiler.txt
83.98
Error: dotty/compiler/: expected 84, but got 83.98 instead
```

ckipp01:
> My guess is that TypeComparer.scala was not parsing correctly, but now that given is parsing correctly it's hitting an unrelated bug (related to if-then).

Makes sense.

ckipp01 (approved):
LGTM!

PR #345, opened 2023-09-07
Sync locals from nvim-tree-sitter.

rwjblue:
1. Add `(block)` as a `@local.scope`
2. Fix the scope for functions. `function_declaration` is used for
   abstract functions, while `function_definition` is used for the
   actual concrete functions (along with their function bodies).
3. Add `(class_parameter)`'s `name` as a local definition.

References:

- https://github.com/nvim-treesitter/nvim-treesitter/pull/4723
- https://github.com/nvim-treesitter/nvim-treesitter/pull/4594

### DIFF
--- a/queries/scala/locals.scm
+++ b/queries/scala/locals.scm
@@ -1,14 +1,17 @@
 (template_body) @local.scope
 (lambda_expression) @local.scope
-
+(block) @local.scope
 
 (function_declaration
-      name: (identifier) @local.definition) @local.scope
-
-(function_definition
       name: (identifier) @local.definition)
 
+(function_definition
+      name: (identifier) @local.definition) @local.scope
+
 (parameter
+  name: (identifier) @local.definition)
+
+(class_parameter
   name: (identifier) @local.definition)
 
 (binding


eed3si9n (approved):
Thanks!

eed3si9n:
```
  ✗ basics.scala
    Failure - row: 79, column: 24, expected highlight 'method', actual highlights: 'none'
```

rwjblue:
Thank you for reviewing, I'll dig into those failures!

rwjblue:
@eed3si9n - As far as I can tell (I'm still ramping up my debugging skills for tree-sitter), the failure here demonstrates a somewhat fundamental issue with how `tree-sitter highlight` actually works. This issue is discussed over in https://github.com/nvim-treesitter/nvim-treesitter/issues/170 (and resolved in https://github.com/nvim-treesitter/nvim-treesitter/pull/295 by allowing function definitions to specify that they are valid in the parent scope as well as their own child scope).

To illustrate the problem I just pushed a test that is being fixed by moving `@local.scope` from `(function_declaration)` to `(function_definition)`.

```scala
def meth_with_params(localParam: Int) {
  var ref_param = c"$localParam $meth_with_params"
//                   ^parameter
//                               ^method
}
var okay1 = s"hello"
var okay = c"$localParam $okay1"
//            ^none
//                        ^variable
```

Specifically, the local function parameter (named `localParam`) should only be in scope within the body of `meth_with_params`, but on master(e.g. without the `@local.scope` move to `(function_definition)`) for
`var okay = c"$localParam $okay1"` the `localParam` is highlighted as a **parameter**!.

Here is the failure for the new test when ran against `master`:

```
    Failure - row: 6, column: 14, expected highlight 'none', actual highlights: 'parameter'
```

rwjblue:
FWIW, I created https://github.com/nvim-treesitter/nvim-treesitter/pull/5359 to ensure that the nvim-treesitter locals for scala are correct. As far as I can tell there is no corresponding feature for `tree-sitter highlight` that we could use. (Please correct me if I'm wrong!!)

PR #347, opened 2023-09-07
Add `(for_expression)` to locals.scm

rwjblue:
This change does two main things:

1. Indicate that `(for_expression)`'s introduce a `@local.scope` (this is the scope for the enumerators to be used within the loop).
2. Indicate that the `(identifiers)` within the `(enumerators)` are `@local.definition`s.

---

For the following example snippet:

```scala
val fruits = List("apple", "banana", "avocado", "papaya")

val countsToFruits = fruits.groupBy(fruit => fruit.count(_ == 'a'))

for ((count, fruits) <- countsToFruits) {
  println(s"with (fruits) 'a' × $count = $fruits")
}
```

The `count` and `fruits` identifiers are new definitions introduced by the `for` expressions scope.

### DIFF
--- a/queries/scala/locals.scm
+++ b/queries/scala/locals.scm
@@ -1,6 +1,6 @@
 (template_body) @local.scope
 (lambda_expression) @local.scope
-
+(for_expression) @local.scope
 
 (function_declaration
       name: (identifier) @local.definition) @local.scope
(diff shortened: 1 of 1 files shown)


ckipp01 (commented):
Hey @rwjblue, same comment as the other PR. Would you be willing to add in some tests in the highlight queries showing this?

Issue #342, opened 2023-08-30
Braceless if-then, function call, return expression


eed3si9n:
### Commit of tree-sitter-scala you tested this on

d50b6ca5cc3d925e3d1f497199cb8d8383ddae8a

### A code sample showing the error

```scala
class C:
  def foo =
    if (aaa) || bbb(ccc) then return true
```

### Show the error node

```
    (compilation_unit
      (class_definition
        (identifier))
      (ERROR
        (identifier)
        (parenthesized_expression
          (identifier))
        (infix_expression
          (operator_identifier)
          (identifier)
          (parenthesized_expression
            (identifier)))
        (identifier)
        (return_expression
          (boolean_literal))))
```

PR #350, opened 2023-09-16
Fix parenthesized expressions in braceless `if` conditions

### LINKED ISSUE #263: Braceless `if` with args in parens and field expressions

374709c

```
object A:
  if (a) && b.c then
    ()
```

### Show the error node

```
(compilation_unit [0, 0] - [3, 0]
  (object_definition [0, 0] - [3, 0]
    body: (template_body [0, 8] - [3, 0]
      (if_expression [1, 2] - [1, 20]
        condition: (parenthesized_expression [1, 5] - [1, 8]
          (identifier [1, 6] - [1, 7]))
        consequence: (postfix_expression [1, 9] - [1, 20]
          (field_expression [1, 9] - [1, 15]
            value: (operator_identifier [1, 9] - [1, 11])
            (ERROR [1, 12] - [1, 13])
            field: (identifier [1, 14] - [1, 15]))
          (identifier [1, 16] - [1, 20])))
      (unit [2, 4] - [2, 6]))))
```

_No response_

[Namer.scala](https://github.com/lampepfl/dotty/blob/f128063530e9e0c22a761470a79103ef5b823dbe/compiler/src/dotty/tools/dotc/typer/Namer.scala#L168)

susliko:
Fixes #263 and #342
`_if_condition` extracted from `if_expression` and given a magic dynamic
precedence of 4


test/highlight/basics.scala
```diff
@@ -26,19 +26,19 @@ object Hello {
   trait Test {
 // ^ keyword 
 //       ^ type
-     def meth(i: Int)(implicit x: Boolean) = ???
+    def meth(i: Int)(implicit x: Boolean) = ???
```
susliko, line 29: I've made definitions in this trait to be of the same indentation as
```scala
trait Foo {
    def meth = ???
  def anonFun = ???
}
```
is parsed with an ERROR and this causes highlighting test to flack

We might want to open an issue on this, but I'm reluctant to doing this as `scalac` raises a warning for this code
```
-- Warning: foo.scala:3:2 ------------------------------------------------------
3 |  def anonFun = ???
  |  ^
  |  Line is indented too far to the left, or a `}` is missing
```

grammar.js
```diff
@@ -1104,22 +1106,25 @@ module.exports = grammar({
...
+        ),
+        field("consequence", $._indentable_expression),
+        optional(seq("else", field("alternative", $._indentable_expression))),
+      )),
+
+    // NOTE(susliko): _if_condition and its magic dynamic precedence were introduced as a fix to
+    // https://github.com/tree-sitter/tree-sitter-scala/issues/263 and
+    // https://github.com/tree-sitter/tree-sitter-scala/issues/342 
+    // Neither do I understand why this works, nor have I found a better solution
+    _if_condition: $ => prec.dynamic(4, choice(
```
susliko, line 1124: I'm ashamed of this

corpus/expressions.txt
```diff
@@ -404,6 +404,11 @@ class C:
     else
       ()
 
+    if (a) || b(c) then return true
```
eed3si9n, line 407: Nice!

eed3si9n (approved):
Thanks @susliko!
I've tried this one in the past and gave up, so happy to see this fix.

PR #369, opened 2023-12-24
Allow infix inline modifier

### LINKED ISSUE #366: inline + infix modifier on methods causes errors

1b4c2fa5c55c5fd83cbb0d2f818f916aba221a42

```scala
inline infix def hello = true
```

Seems to cause this beautiful highlighting 
CleanShot 2023-10-18 at 16 03 40

### Show the error node

```
  (ERROR [0, 0] - [0, 29]
    (inline_modifier [0, 0] - [0, 6])
    (identifier [0, 7] - [0, 12])
    (identifier [0, 13] - [0, 16])
    (assignment_expression [0, 17] - [0, 29]
      left: (identifier [0, 17] - [0, 22])
      right: (boolean_literal [0, 25] - [0, 29]))))
```

single function declaration with 2 modifiers

everywhere, reproducible from this repo alone

susliko:
Fixes #366

Rebalance precedences of `match_expression` (which can contain "inline") and `modifiers`

### DIFF
--- a/corpus/definitions.txt
+++ b/corpus/definitions.txt
@@ -1661,8 +1661,7 @@ Infix methods (Scala 3)
 ================================================================================
 
 object Test:
-  infix private def hello = 25
-
+  inline infix private def hello = 25
 --------------------------------------------------------------------------------
 
 (compilation_unit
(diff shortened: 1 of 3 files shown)


eed3si9n:
> Error: scala_scala/src/library/: expected 100, but got 99.56 instead

apparently a regression on parsing scala-library.

eed3si9n (approved):
Thanks!

Issue #173, opened 2023-01-20
Unable to use parser for indentation with significant whitespace syntax


ckipp01:
### Commit of tree-sitter-scala you tested this on

97d5e7c86492cb4449e6f7e1b8577e20b4bfb046

### A code sample showing the error

```scala
object Main:
```

### Show the error node

So this is sort of a discussion, but also something that I think we'll need to change if we want to support indents. The issue is that in the above code sample the tree at that point would look like this:

```
object_definition [0, 0] - [0, 11]
  "object" [0, 0] - [0, 6]
  name: identifier [0, 7] - [0, 11]
ERROR [0, 11] - [2, 0]
  ":" [0, 11] - [0, 12]
```

If your cursor is here:

```scala
object Main:<cursor>
```
And you hit `<enter>` then you'd expect it to be indented. An indent query for this would look like this:

```
[
  (object_definition)
] @indent
```

From the [`nvim-treesitter`](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#indents) docs it says:

> @indent         ; indent children when matching this node

The project is that this won't match here since the `ERROR` node isn't a child of the `object_definition` and `:` is actually a child of the current `ERROR` node.

Realistically even if you have an opening `{` you have the same issue:

```
object_definition [0, 0] - [0, 11]
  "object" [0, 0] - [0, 6]
  name: identifier [0, 7] - [0, 11]
ERROR [0, 12] - [0, 13]
  "{" [0, 12] - [0, 13]
```

But since the vast majority of editors will give you a closing `}` by default it will really look like this:

```
object_definition [0, 0] - [0, 15]
  "object" [0, 0] - [0, 6]
  name: identifier [0, 7] - [0, 11]
  body: template_body [0, 12] - [0, 15]
    "{" [0, 12] - [0, 13]
    "}" [0, 14] - [0, 15]
```

Meaning that you're in `object_definition`, so the `<enter>` will indent like you'd expect.

### What do you expect the tree to look like

I'd expect that when you have the following:

```scala
object Main:<cursor>
```
And I hit `<enter>` there is a way that we can indent correctly. 

### Where are you experiencing this error?

nvim-treesitter

susliko:
@ckipp01 🖖🏻 
I haven't found a way to influence the position of ERROR node in the tree. 
However, we can write a query for what we have now, i.e. match the ERROR node! The following one seems to solve the described issue:
```scheme
[...]

PR #378, opened 2024-03-10
Nvim indents

susliko:
Sharing queries for proper indents (also fixes problem descirbed in #173).
I'm going to use them for a while and polish before merging

cc @keynmol


susliko:
TODO
```scala
object Foo:
  class Bar:<cr>
```
---

farnoy:
Hey @susliko , how is your experience so far with this change? If it's not regressing anything else, maybe it may be worth merging to have an incremental improvement?

susliko:
I've been using these indentation queries for neovim for a while now and I find them to be more or less satisfactory.

eed3si9n (approved):
Thanks!

PR #387, opened 2024-03-12
Update tree-sitter to 0.21.0

eed3si9n:
I originally sent a PR to bump to tree-sitter 0.22.1, but maybe that's going too ahead.
Here's a PR to bump to tree-sitter 0.21.0.

Ref https://github.com/neovim/neovim/pull/27566

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "tree-sitter-scala"
 description = "scala grammar for the tree-sitter parsing library"
-version = "0.22.0"
+version = "0.21.0"
 keywords = ["incremental", "parsing", "scala"]
 categories = ["parsing", "text-editors"]
 repository = "https://github.com/tree-sitter/tree-sitter-scala"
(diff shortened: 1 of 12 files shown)


Cargo.toml
```diff
@@ -1,11 +1,11 @@
 [package]
 name = "tree-sitter-scala"
 description = "scala grammar for the tree-sitter parsing library"
-version = "0.20.0"
+version = "0.22.0"
```
ckipp01, line 4: Should this be `0.21.0`?

package.json
```diff
@@ -3,23 +3,44 @@
...
+  ],
   "author": "Max Brunsfeld",
   "license": "MIT",
   "dependencies": {
-    "nan": "^2.14.1"
+    "node-addon-api": "^7.1.0",
+    "node-gyp-build": "^4.8.0"
+  },
+  "peerDependencies": {
+    "tree-sitter": "^0.21.0"
```
ckipp01, line 26: Do we want to lock it at `0.21.0` instead of using `^`?

Cargo.toml
eed3si9n: Addressed in Addressed in 4c7275c70e41c7c1fad6f4e52a60dd3941db5bba

ckipp01 (approved):
LGTM!

Issue #392, opened 2024-04-12
Parser gets stuck


Moddus:
### Commit of tree-sitter-scala you tested this on

70b4fe63c4

### A code sample showing the error

Hi,

It looks like the the grammar on commit `70b4fe63c4` is causing an infinite loop e.g. running within `tree-sitter-scala`:
````
tree-sitter parse sample.scala --timeout 1000000
````

Please see attached the sample.scala:
[sample.scala.txt](https://github.com/tree-sitter/tree-sitter-scala/files/14962601/sample.scala.txt)

Could someone help explaining how this can be debugged?

Thanks for your time and suggestions

### Show the error node

No error message as execution of `tree-sitter` is stuck.

### Where are you experiencing this error?

Running tree-sitter tool in CLI.

eed3si9n:
Maybe try binary search to minimize the example, and/or git bisect to find out when this has occurred.
If you go back to earlier tags in https://github.com/tree-sitter/tree-sitter-scala/releases does the problem still persist?

Moddus:
Good points! I can confirm `v0.20.2` works where `v0.20.3` and `v0.21.0` don't.

ObserverOfTime:
Parsing with `v0.20.2` reveals that there's a `}` missing at the end of the file.
Adding it allows it to be parsed with master (though there's another error).

Bisecting points to ef639d4 and consequently to 045ce59.

eed3si9n:
@ObserverOfTime Thanks!
So the relevant PR is https://github.com/tree-sitter/tree-sitter-scala/pull/350 where `_if_condition` precedence became dynamic.

susliko:
The issue was fixed by #524 

I don't know how, though..

PR #393, opened 2024-04-13
Replace dynamic prec in if-then with an additional node

eed3si9n:
## Problem
#350 introduced a dynamic precedence to resolve the conflict of Scala 2 and Scala 3 grammar for if-then.

## Solution
This removes the dynamic precedence and replaces it with prec.right around `then` side.

## Note
There's a report of parser getting stuck #392 and and I was hoping that removing dynamic precedence would fix that, but it didn't seem to.


eed3si9n:
The regression in smoke test is on BigInt.scala, which kind of looks like the following:

```scala
class A
{
  def bar(that: A): A =
    if (true) A(1)
    else 2

  def foo(that: A): (A, A) =
    if (true) 1
    else 2
}
```

I think it fails to parse the above because it thinks `(A, A) =
    if (true) 1
    else 2` is an infixed type, which is a separate issue we probably need to address.

susliko:
Just a note.

Running `tree-sitter parse -d sample.scala.txt` gives the line where parser stucks - it's line 969.
If we remove a newline between lines 968 and 969 it stops getting stuck.

For now I had no luck in fixing this.

PR #394, opened 2024-04-14
Exclude colon, at, and equal from single opchar

eed3si9n:
## Problem
Currently `operator_identifier` includes characters like `:` (colon), `@` (at), and `=` (equal) even though they cannot be a legal operator without backticks. Having equal etc pushes tree-sitter into falsely thinking some construct to be an infix operation when they are `=`.

Another complication is Unicode [Math symbols](https://www.compart.com/en/unicode/category/Sm), which includes equal sign.

## Solution
1. Remove colon, at, equal sign, and Math symbols from the single-char `operator_identifier`.
2. Add back back a few Math symbol unicodes.

All tests pass as-is.

## Note

```scala
$ scala
Welcome to Scala 3.4.0 (1.8.0_402, Java OpenJDK 64-Bit Server VM).
Type in expressions for evaluation. Or try :help.

scala> def : = 0
-- [E040] Syntax Error: --------------------------------------------------------
1 |def : = 0
  |    ^
  |    an identifier expected, but ':' found
  |
  | longer explanation available when compiling with `-explain`

scala> def @ = 0
-- [E040] Syntax Error: --------------------------------------------------------
1 |def @ = 0
  |    ^
  |    an identifier expected, but '@' found
  |
  | longer explanation available when compiling with `-explain`

scala> def = = 0
-- [E040] Syntax Error: --------------------------------------------------------
1 |def = = 0
  |    ^
  |    an identifier expected, but '=' found
  |
  | longer explanation available when compiling with `-explain`
```


script/smoke_test.sh
```diff
@@ -3,7 +3,7 @@
 # This is an integration test to generally check the quality of parsing.
 
 SCALA_SCALA_LIBRARY_EXPECTED=100
-SCALA_SCALA_COMPILER_EXPECTED=96
+SCALA_SCALA_COMPILER_EXPECTED=97
```
eed3si9n, line 6:
```
Notice: ok, scala_scala/src/compiler/: 97.54%, expected at least 96%
```

ckipp01 (approved):
Nice improvement!

PR #398, opened 2024-05-10
Updates

### LINKED ISSUE #344: Need help getting access to NPM (please help us if you see this max :) )
The latest release [appears to be 0.20.2](https://github.com/tree-sitter/tree-sitter-scala/releases/tag/v0.20.2) however the latest published package on npm is [0.19.0](https://www.npmjs.com/package/tree-sitter-scala).

Is NPM an official release channel for this package?  If so could we have the latest version published there.  Thanks.

### LINKED ISSUE #384: Update to tree-sitter 0.22.x?
We're behind on the tree-sitter version, so we should consider bumping the version up for code gen and testing purpose.

## breaking changes in 0.21.0
https://github.com/tree-sitter/tree-sitter/releases/tag/v0.21.0

> - Remove the apply-all-captures flag, make last-wins precedence the default
>
>    NOTE: This change might cause breakage in your grammar's highlight tests.
Just flip the order around of the relevant queries, and keep in mind that the
last query that matches will win.

The highlighting became last-wins in 0.21.0.

## other parsers

- https://github.com/tree-sitter/tree-sitter-typescript (0.20.6)
- https://github.com/tree-sitter/tree-sitter-javascript (0.20.4)
- https://github.com/tree-sitter/tree-sitter-rust (0.20.4)
- https://github.com/tree-sitter/tree-sitter-python (0.21.0)
- https://github.com/tree-sitter/tree-sitter-c (0.21.0)

amaanq:
Closes #344
Closes #384

this is the first part of some changes/fixes I'll make to this repo (the forthcoming ones being more grammar/scanner bug fixes) - I know this one has several contributors/more activity compared to the other upstream grammars (where it's pretty much just me maintaining it ;) ), so I'll leave this up for a bit in case anyone has questions/comments

I'll also be updating the repo secrets so we can get this grammmar published to npm and pypi :grin: I think the current CI is a little overkill - all we need to do is test that is parses the test corpus, scala3, and scala, and I've dumped all known failures into a file, so regressions can be found easily if a file outside of *these* fails to parse correctly


.github/workflows/ci.yml
```diff
@@ -1,116 +1,50 @@
...
     steps:
-      - name: checkout tree-sitter-scala
+      - name: Checkout repository
         uses: actions/checkout@v4
+      - name: Set up tree-sitter
+        uses: tree-sitter/setup-action/cli@v1
+      - name: Set up examples
+        run: |-
+          git clone https://github.com/scala/scala examples/scala --single-branch --depth=1 --filter=blob:none
+          git clone https://github.com/scala/scala3 examples/scala3 --single-branch --depth=1 --filter=blob:none
```
eed3si9n, line 40: We probably want to pin this to a specific version. Currently we use 2.13.10 and 3.3.0 for example.

script/smoke_test.sh
eed3si9n, line 1: Why was this removed again? It's useful for us know the parsing % of the example repos when we review PRs.

.github/workflows/sync.yml
eed3si9n, line 1: Why is this getting removed? I feel like humans checking in generated C code is inviting for security issues. This at least lets us automate the C codegen.

script/smoke_test.sh
amaanq: The ci action should take care of it, rn it will error out if any file that's *not* excluded fails, but I can def add this back for local devs to see what % passes if that's useful here

.github/workflows/sync.yml
amaanq: I personally don't think this is a good idea *unless* it always uses the latest version of tree-sitter. The generate process can and does change output from version to version, so it's always good to use the latest version. That being said, we also plan to get rid of the need to check in parser.c in the near future, so this won't be necessary then. I can add this back if it provides value for you all, but I think checking to see if parser.h or other files were unnecessarily changed due to using an older version as well as eyeballing a few lines that *did* change is okay enough (again I know infra here sucks, this is what I have to do across dozens of parsers, so I don't mind adding it back if it is useful to you)

eed3si9n: > I personally don't think this is a good idea unless it always uses the latest version of tree-sitter.

Assuming that C code won't  be forward compatible, wouldn't the version of tree-sitter runtime used by major editors dictate the rollout timing? (Not sure what happens if Neovim uses version X and we generate X + 1)

amaanq: tree-sitter's library has shims in place to ensure backwards compatibility for a while (in the case of *breaking* changes), but more of what I meant was there have been improvements to the parser compilation time w.r.t how large character sets are lexed. Any consumer can compile the parser generated before or after this and use the parser with the same tree-sitter library version with no problem at all, but obviously we'd want the newer version with these optimizations

but eventually, let's take for example the addition of a name field to the language struct, that won't be breaking if we generate before or after that because the library's runtime itself has checks for the ABI version anyways to see if the field is present or not

amaanq:
well, some good news is the scanner stack changes fixed *some* edit fuzzing bugs :partying_face: was trying to replicate one and I couldn't, but I managed to get another

script/smoke_test.sh
ObserverOfTime: (In which case you should make the step `continue-on-error`)

PR #402, opened 2024-06-30
Move corpus

eed3si9n:
**Problem/Solution**
I noticed that corpus tests weren't running.
Apparently tree-sitter 0.22.0 moved the directory -https://github.com/tree-sitter/tree-sitter/releases/tag/v0.22.0


eed3si9n:
tree-sitter_tree-sitter-scala_a6e8f24

PR #403, opened 2024-07-03
Fix for https://github.com/fluencelabs/aqua/blob/main/parser/src/main…

potan:
Supporting blocks with "{}" in "given" definitions.
```scala
object {
  given g: Show[T] with {
    def show(x: T) = ""
  }
}
```


susliko:
@potan Could you please provide a description for an issue addressed by this PR and a minimal piece of code which fails to be parsed by tree-sitter-scala parser?

exoego:
Thanks for the PR.
Superseded by https://github.com/tree-sitter/tree-sitter-scala/pull/666

PR #404, opened 2024-07-17
stack: prevent undefined behavior on int pointers

psarna:
Without proper alignment, UBSAN will cause the program to fail at runtime with "illegal instruction" signal. `zig cc` toolchain enables UBSAN by default, which helped me pinpoint the cause.

### DIFF
--- a/src/stack.h
+++ b/src/stack.h
@@ -1,5 +1,6 @@
 #include <stdbool.h>
 #include <stdio.h>
+#include <stdint.h>
 #include <stdlib.h>
 #include <string.h>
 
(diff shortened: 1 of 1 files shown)


eed3si9n:
There was a commit https://github.com/tree-sitter/tree-sitter-scala/pull/398/commits/2c5c1d2f86ad9ba9e0604174c9cf06334f54b7fe in #398, which I wonder we should cherry pick instead.

eed3si9n:
Closing this in favor of https://github.com/tree-sitter/tree-sitter-scala/pull/405

PR #405, opened 2024-07-17
refactor(scanner): use new array header for stack

eed3si9n:
## Problem

@psarna wrote in #404:

> Without proper alignment, UBSAN will cause the program to fail at runtime with "illegal instruction" signal. 

## Solution

This cherry picks scanner change https://github.com/tree-sitter/tree-sitter-scala/commit/2c5c1d2f86ad9ba9e0604174c9cf06334f54b7fe from https://github.com/tree-sitter/tree-sitter-scala/pull/398.


ckipp01 (approved):

amaanq:
thank you for cherry picking my work @eed3si9n, I intended to make it easier to review/split it up, but didn't have time. and thanks for the NPM releasing too!

PR #408, opened 2024-08-05
ci: Add GitHub Actions to publish

eed3si9n:
Fixes https://github.com/tree-sitter/tree-sitter-scala/issues/344

## Problem

NPM packages are not published.

## Solution

This cherry picks GitHub Action releasing from https://github.com/tree-sitter/tree-sitter-scala/pull/398

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,23 @@
+name: Publish package
+
+on:
+  push:
+    tags: ["*"]
+
+concurrency:
+  group: ${{github.workflow}}-${{github.ref}}
+  cancel-in-progress: true
+
+jobs:
+  npm:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,24 @@
...
+concurrency:
+  group: ${{github.workflow}}-${{github.ref}}
+  cancel-in-progress: true
+
+jobs:
+  npm:
+    uses: tree-sitter/workflows/.github/workflows/package-npm.yml@main
+    secrets:
+      NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
+  # TODO: comment this out for now since we publish the crate manually
```
eed3si9n, line 16: https://crates.io/crates/tree-sitter-scala

PR #409, opened 2024-08-07
Allow literal types to be operands of an infix type

### LINKED ISSUE #406: Highlighting fails for string literal unions

45b5ba0e749a8477a8fd2666f082f352859bdc3c

Syntax highlighting for string literal union types fails under multiple circumstances. I first noticed this on Scastie (issue posted [here](https://github.com/scalacenter/scastie/issues/1088)), but also experience it on Zed.

Here is a screenshot of the failing cases I've found so far

image

```scala
// works!
type test1 = "a" | "b"
```
```scala
// fails after two elements
type test2 = "a" | "b" | "c"
```
```scala
// breaks from tokens on next line
type test3 = "a" | "b"
// like this comment
```
```scala
type test4 = "a" | "b"
type test4_ = "x" | "y"
// breaks subsequent lines as well
```
```scala
// fails for parameter types
def test5(x: "a" | "b") = ???
```
```scala
// fails for return types
def test6(): "a" | "b" = ???
```

Zed version: 0.142.6
Zed Scala extension version: 0.0.3

### Show the error node

will add later if I have time to figure out. I am not familiar with how tree-sitters work.

_No response_

Zed, Scastie

susliko:
Fixes #406

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -875,7 +875,7 @@ module.exports = grammar({
     _infix_type_choice: $ =>
       prec.left(
         PREC.infix,
-        choice($.compound_type, $.infix_type, $._annotated_type),
+        choice($.compound_type, $.infix_type, $._annotated_type, $.literal_type),
       ),
 
     infix_type: $ =>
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Thanks!

PR #422, opened 2024-09-02
build: update bindings

amaanq:


amaanq:
@eed3si9n I didn't know Max wasn't apart of the crate owners for the Rust package for this grammar - do you mind inviting him or me so we can turn the publishing action [green](https://github.com/tree-sitter/tree-sitter-scala/actions/runs/10665038626/job/29577989934)? Now all that a maintainer here needs to do is tag a commit & push the tag and the CI job will publish each package across python/js/rust

eed3si9n:
I'm pretty sure I did invite you both a while back, but let me invite again.

eed3si9n:
Also was a reason this and https://github.com/tree-sitter/tree-sitter-scala/commit/a366c0d52485b84d2d9ba66dee530398526acb01 were merged without review and while the tests failed?
With tree-sitter-scala, first we land a PR, we run a GitHub Action to securely generate the C code, and then tag the release with the C code in the master branch.

PR #426, opened 2024-09-10
Fix type alias violation in scanner

yangchi:
The cast from buffer which is char* to uint16_t* is a type alias violation.
If tree-sitter-scala is build with UBSAN enabled this is a runtime error.
This PR fixes it by replacing these cast + assignment with just plain old memcpy.

### DIFF
--- a/src/scanner.c
+++ b/src/scanner.c
@@ -59,15 +59,15 @@ unsigned tree_sitter_scala_external_scanner_serialize(void *payload, char *buffe
   }
 
   size_t size = 0;
-  *(int16_t *)&buffer[size] = scanner->last_indentation_size;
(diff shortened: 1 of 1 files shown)


eed3si9n (approved):
Thanks!

yangchi:
I'm getting the following error in windows-latest test on GH actions:
```
Could not find 'D:\a\tree-sitter-scala\tree-sitter-scala\bindings\node\*_test.js'
```
Did I miss some steps to generate some files, or is this unrelated to my change? If it's the latter, is there an easy way for me to rerun the checks? Thank you!

yangchi:
I see. I can wait for #425 to go in first. Thank you!

PR #431, opened 2024-10-15
Update bindings for 0.24

amaanq:
This PR has a few changes stemming from upstream, namely:
 - migrate to `tree-sitter.json`
 - tweak language bindings
 - use `github` workflow for uploading artifacts on release


eed3si9n:
Could you try removing any generated files from the PR and let the GitHub Action generate it plz? I think it's checking for the fidelity of all files because it detected a new .c file.

amaanq:
> Could you try removing any generated files from the PR and let the GitHub Action generate it plz? I think it's checking for the fidelity of all files because it detected a new .c file.

I didn't add a c file though, let alone any generated ones. I question the fidelity script here (and this is the only grammar repo I know of that uses this)

bindings/swift/TreeSitterScala/scala.h
eed3si9n, line 1: Maybe this file?

amaanq: It's a swift header file...

eed3si9n: Yea, currently the sync job is setup to generate and commit bindings file. I think the YAML should be updated to track the expanded sets up binding files:
https://github.com/tree-sitter/tree-sitter-scala/blob/master/.github/workflows/sync.yml

amaanq: Please don't do that. Bindings require manual intervention to fix missing information that can't be automatically assumed, like the scanner file info and query files.

eed3si9n: We can relax it for Swing binding file etc, but tree-sitter-scala currently does not manually edit the bindings files as far as I know. Doesn't tree-sitter CLI regularly overwrite these?

amaanq: Nah, it got moved to the `init` command, because this became problematic and was doing too much

eed3si9n: ok here's https://github.com/tree-sitter/tree-sitter-scala/pull/432

eed3si9n (approved):

PR #432, opened 2024-10-16
Relax the fidelity check for bindings

eed3si9n:
## Problem
Apparently tree-sitter bindings are expected to beccreated manually and are allowed to be hand-edited. Currently it triggers fidelity check.

## Solution
This relaxes the C code fidelity check so bindings can be updated.


amaanq (approved):

keynmol (approved):

amaanq:
@eed3si9n I totally forgot, but I did get the invitation to be an owner of this package on crates.io, but the invitation expired. Can I get another invite? Sorry about that

eed3si9n:
@amaanq Sent.

amaanq:
Nice, one step closer to removing `parser.c` from grammar repos - https://github.com/tree-sitter/tree-sitter-scala/actions/runs/11772732014

Issue #264, opened 2023-05-28
Braceless `match` with cases of same indentation


susliko:
### Commit of tree-sitter-scala you tested this on

374709c

### A code sample showing the error

```scala
object A:
  a match
  case b =>
  case c =>
```

### Show the error node

```
(compilation_unit [0, 0] - [4, 0]
  (object_definition [0, 0] - [4, 0]
    name: (identifier [0, 7] - [0, 8])
    body: (template_body [0, 8] - [4, 0]
      (ERROR [1, 2] - [2, 6]
        (identifier [1, 2] - [1, 3]))
      (identifier [2, 7] - [2, 8])
      (ERROR [2, 9] - [3, 11]))))
```

### Where are you experiencing this error?

[Applications.scala](https://github.com/lampepfl/dotty/blob/f128063530e9e0c22a761470a79103ef5b823dbe/compiler/src/dotty/tools/dotc/typer/Applications.scala#L118)

PR #436, opened 2024-11-25
feat: support non-indented cases in Scala3 match expression

susliko:
Addresses #264

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -95,6 +95,8 @@ module.exports = grammar({
     [$._if_condition, $._simple_expression],
     // _postfix_expression_choice  ':'  '('  wildcard  •  ':'  …
     [$.binding, $._simple_type],
+    // expression  'match'  'case'  _case_pattern  ';'  _automatic_semicolon  expression  •  'match'  …
+    [$.match_expression, $._block],
   ],
 
   word: $ => $._alpha_identifier,
@@ -806,6 +808,9 @@ module.exports = grammar({
     indented_cases: $ =>
       prec.left(seq($._indent, repeat1($.case_clause), $._outdent)),
 
+    non_indented_cases: $ =>
+      prec.right(sep1($._automatic_semicolon, $.case_clause)),
+
     _indented_type_cases: $ =>
       prec.left(seq($._indent, repeat1($.type_case_clause), $._outdent)),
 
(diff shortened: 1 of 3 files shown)


grammar.js
```diff
@@ -1157,7 +1162,7 @@ module.exports = grammar({
         optional($.inline_modifier),
         field("value", $.expression),
         "match",
-        field("body", choice($.case_block, $.indented_cases)),
+        field("body", choice($.case_block, $.indented_cases, $.non_indented_cases)),
```
susliko, line 1165: Should `non_indented_cases` be aliased to `indented_cases` here to preserve compatibility for any queries that may not expect nodes other than `case_block`/`indented_cases`, wdyt?

susliko:
Looks like CI ran against an old version, without changes in grammar.js. Strange 🤔

test/corpus/expressions.txt
```diff
@@ -850,8 +866,8 @@ class C {
     a
       .b
       // comment1
-      /* 
-         comment2 
+      /*
```
susliko, line 869: Changes here and below are caused by changes in behavior of `tree-sitter test -u`, which now removes trailing spaces

grammar.js
eed3si9n: +1 to this.

eed3si9n:
> Looks like CI ran against an old version, without changes in grammar.js. Strange 🤔

Previously, we were running generate explicitly only on Linux, but at some point CI started testing on macOS as well? I fixed the CI in https://github.com/tree-sitter/tree-sitter-scala/pull/437

eed3si9n (approved):
LGTM

eed3si9n:
```
Checking syntax complexity: npm exec -c 'tree-sitter generate --report-states-for-rule compilation_unit' 2>&1 >/dev/null
Report written to report-complexity.txt
tuple_type 	1857
Error: complexity of the most complex definition tuple_type: 	1857, higher than the allowed ceiling 1400
```

I guess PREC.right adds to the complexity a bit? I'm ok with bumping this up.

susliko:
While working on this PR, I realized that the conflict between `$.match_expression` and `$._block` is slowing down the generation process by about 9 times. I haven't found a good solution yet, so I'm going to close the PR for now.

Issue #438, opened 2024-12-12
SIP-47: Clause Interleaving


eed3si9n:
### Commit of tree-sitter-scala you tested this on

5f44942205c2364ce2ced14a40687d1e09685034

### A code sample showing the error

```scala
  // SIP-47 - Clause Interleaving
  def getOrElse(k: Key)[V >: k.Value](default: V): V
```

### Show the error node

```lisp
          (function_declaration
            (identifier)
            (ERROR
              (parameters
                (parameter
                  (identifier)
                  (type_identifier))))
            (type_parameters
              (identifier)
              (lower_bound
                (stable_type_identifier
                  (identifier)
                  (type_identifier))))
            (parameters
              (parameter
                (identifier)
                (type_identifier)))
            (type_identifier)))))
```

PR #439, opened 2024-12-12
SIP-47 Clause Interleaving support

eed3si9n:
Fixes https://github.com/tree-sitter/tree-sitter-scala/issues/438

## Problem
Scala 3.6.2 added interleaving params <https://docs.scala-lang.org/sips/clause-interleaving.html>,
as in term parameters, type parameters, and
using parameters can show up in an arbitrary order.

## Solution
This adds support for it.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -550,10 +550,14 @@ module.exports = grammar({
       prec.right(
         seq(
           field("name", $._identifier),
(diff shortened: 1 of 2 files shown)


susliko (approved):
LGTM

PR #443, opened 2024-12-12
test: Add test for SIP-62 for

eed3si9n:
## Problem/Solution
This add a test for [SIP-62](https://docs.scala-lang.org/sips/better-fors.html), which allows for comprehension that starts with an alias.


susliko (approved):
Nice

PR #446, opened 2024-12-14
SIP-58 Named tuples support

eed3si9n:
## Problem

[SIP-58 Named tuples](https://docs3.scala-lang.org/sips/named-tuples.html) are now available as an experimental feature.

## Solution
This implements a support for them.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -91,6 +91,11 @@ module.exports = grammar({
     [$.class_parameters],
     // 'for'  operator_identifier  ':'  _annotated_type  •  ':'  …
     [$._type, $.compound_type],
+    // 'given'  '('  operator_identifier  ':'  _type  •  ','  …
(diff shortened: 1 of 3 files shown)


grammar.js
```diff
@@ -991,6 +1016,7 @@ module.exports = grammar({
         $.interpolated_string_expression,
         $.capture_pattern,
         $.tuple_pattern,
+        $.named_tuple_pattern,
         $.case_class_pattern,
```
keynmol, line 1020: I think we also need a modified case_class_pattern with the named arguments: https://docs.scala-lang.org/sips/named-tuples.html#pattern-matching-with-named-fields-in-general

eed3si9n: Addressed in 690838e56c20c2ab1ccaa033a54cc6ee8a787164

keynmol (approved):

PR #456, opened 2025-03-04
Update CONTRIBUTING.md

jonshea:
Some commands previous described in `CONTRIBUTING.md` have been removed, and some of the tree-sitter documentation links are broken. This PR updates `CONTRIBUTING.md` to fix these issues.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,7 +36,7 @@ to `grammar.js`, regenerate and recompile the parser, and run the tests:
 
 ```sh
 tree-sitter generate
-npm test
+tree-sitter test
 ```
 
 Then adjust as necessary. Note that depending on your change you may also have
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,7 +51,7 @@ means much of our testing is in relation to it. You can find the syntax
 highlighting tests in `test/highlight/*.scala`. You can read more about this
 type of testing 
 [here](https://tree-sitter.github.io/tree-sitter/3-syntax-highlighting.html#unit-testing). 
-These test will be run automatically with `npm run test`.
+These test will be run automatically with `tree-sitter test`.
 
 ### tree-sitter highlight
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,11 +7,12 @@ tree-sitter.
 
 ## Requirements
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -26,15 +27,15 @@ project's dependencies:
...
 find details on how testing works with tree-sitter
-[here](https://tree-sitter.github.io/tree-sitter/creating-parsers#command-test).
+[here](https://tree-sitter.github.io/tree-sitter/creating-parsers/5-writing-tests.html).
 
 Once you've added your test case you'll want to then make the required changes
 to `grammar.js`, regenerate and recompile the parser, and run the tests:
 
 ```sh
-npm run build
+tree-sitter generate
```
jonshea, line 38: Is this the command to use now? I see it is used in CI (https://github.com/tree-sitter/tree-sitter-scala/blob/master/.github/workflows/sync.yml#L27), but I’m not sure if anything else also needs to be called?

CONTRIBUTING.md
```diff
@@ -48,15 +49,18 @@ Right now the most common use-case for syntax highlight using tree-sitter is
 [nvim-treesitter](https://github.com/nvim-treesitter/nvim-treesitter), which
 means much of our testing is in relation to it. You can find the syntax
 highlighting tests in `test/highlight/*.scala`. You can read more about this
-type of testing
-[here](https://tree-sitter.github.io/tree-sitter/syntax-highlighting#unit-testing). These test will be automatically ran with `npm run test`.
+type of testing 
+[here](https://tree-sitter.github.io/tree-sitter/3-syntax-highlighting.html#unit-testing). 
+These test will be run automatically with `npm run test`.
```
jonshea, line 54: Is this still true? I only see `node --test bindings/node/*_test.js` in (package.json)[https://github.com/tree-sitter/tree-sitter-scala/blob/master/package.json#L50C14-L50C49]. CI seems like it has different actions for parser and binding tests.

CONTRIBUTING.md
eed3si9n: I think the main testing CI script is https://github.com/tree-sitter/parser-test-action/blob/master/action.yml, and looks like it should be

```bash
tree-sitter generate
tree-sitter test
```

I think.

CONTRIBUTING.md
eed3si9n: I think this is also `tree-sitter test`.

eed3si9n (approved):

PR #459, opened 2025-03-17
Support for string escapes

jonshea:
I also slightly changed the definition of `interpolated_string` by
qualifying the open quotes with `token.immediate(…)`. Prior to this
change the rule would incorrectly match `foo ""` as an `interpolated_string_expression`

I ran these changes against all of the `.scala` files in
https://github.com/scala/scala3 and https://github.com/scala/scala.

The files in `scala/scala3` that newly have errors are:
* `tests/neg/fEscapes.scala`
* `tests/neg/unicodeEscapes-interpolations.scala`
* `tests/pos/multiLineOps.scala`
* `tests/run/i14164.scala`

The first two are tests containing examples of invalid escape
sequences that are expected to fail. `multiLineOps.scala` contains a
line `send_! "!"` that now parses with an error. Previously this
parsed as an `interpolated_string_expression`, which is also entirely
incorrect, So this error is a result of the adding
`token.immediate('"')` to the definition of `interpolated_string`, and
I do not think the change is a regression. Similarly, `i14164.scala`
contains a multi-line expression that previously incorrectly parsed to
`interpolated_string_expression`, and now parses more correctly,
though with an error.

The files in `scala/scala` that newly have errors are similar. Two
test files with intentionally broken escape sequences, and the same
`multiLineOps.scala`.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -32,19 +32,24 @@ module.exports = grammar({
   externals: $ => [
     $._automatic_semicolon,
     $._indent,
-    $._interpolated_string_middle,
-    $._interpolated_string_end,
-    $._interpolated_multiline_string_middle,
-    $._interpolated_multiline_string_end,
     $._outdent,
-    $._simple_multiline_string,
-    $._simple_string,
+    $._simple_string_start,
+    $._simple_string_middle,
+    $._simple_multiline_string_start,
+    $._interpolated_string_middle,
+    $._interpolated_multiline_string_middle,
+    $._raw_string_start,
+    $._raw_string_middle,
+    $._raw_string_multiline_middle,
+    $._single_line_string_end,
+    $._multiline_string_end,
     "else",
     "catch",
     "finally",
     "extends",
     "derives",
     "with",
+    $.error_sentinel,
   ],
 
   inline: $ => [
@@ -209,7 +214,7 @@ module.exports = grammar({
           "package",
           field("name", $.package_identifier),
           // This is slightly more permissive than the EBNF in that it allows any
-          // kind of delcaration inside of the package blocks. As we're more
+          // kind of declaration inside of the package blocks. As we're more
           // concerned with the structure rather than the validity of the program
           // we'll allow it.
           field("body", optional($.template_body)),
@@ -677,7 +682,7 @@ module.exports = grammar({
           // In theory structural_type should just be added to simple_type,
           // but doing so increases the state of template_body to 4000
           $._structural_type,
-          // This adds _simple_type, but not the above intentionall/y.
+          // This adds _simple_type, but not the above intentionally.
           seq($._simple_type, field("arguments", $.arguments)),
           seq($._annotated_type, field("arguments", $.arguments)),
           seq($.compound_type, field("arguments", $.arguments)),
@@ -1540,14 +1545,14 @@ module.exports = grammar({
 
     /**
      * Regex patterns created to avoid matching // comments and /* comment starts.
-     * This could technically match illeagal tokens such as val ?// = 1
+     * This could technically match illegal tokens such as val ?// = 1
      */
     operator_identifier: $ =>
       token(
         choice(
           // opchar minus colon, equal, at
           // Technically speaking, Sm (Math symbols https://www.compart.com/en/unicode/category/Sm)
-          // should be allowed as a single-characeter opchar, however, it includes `=`,
+          // should be allowed as a single-character opchar, however, it includes `=`,
           // so we should to avoid that to prevent bad parsing of `=` as infix term or type.
           /[\-!#%&*+\/\\<>?\u005e\u007c~\u00ac\u00b1\u00d7\u00f7\u2190-\u2194\p{So}]/,
           seq(
@@ -1616,7 +1621,7 @@ module.exports = grammar({
             choice(
               seq(
                 "\\",
-                choice(/[^xu]/, /uu?[0-9a-fA-F]{4}/, /x[0-9a-fA-F]{2}/),
+                choice(/[^xu]/, /[uU]+[0-9a-fA-F]{4}/, /x[0-9a-fA-F]{2}/),
               ),
               /[^\\'\n]/,
             ),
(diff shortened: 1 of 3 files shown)


jonshea:
This is my first time working with tree-sitter, so I would appreciate any feedback. I am honestly not certain this feature warrants the complexity of these changes, so it will not hurt my feelings if we ultimately decide not to merge this PR. But if people do want it, I am happy to iterate on any necessary changes.

I split the external tokens for `SIMPLE_STRING` to have a start / middle / end, and `SIMPLE_MULTILINE_STRING` to have a start / end. As far as I can tell, there are no escape or other tokens that can be found in a simple multiline string, so it could have stayed as a single token, but I think the code is cleaner if we treat it like the other string types. I also added a raw string and a raw multiline string type. I believe the rules for raw multiline strings are identical to the rules for interpolated multiline strings, so we could merge those two nodes if desired.

This implementation uses the same node name (`interpolated_string_expression`) for raw strings and interpolated strings, and the same name (`escape_sequence`) for both escape sequences and dollar sign escapes. Those were arbitrary choices on my part.

grammar.js
```diff
@@ -32,19 +32,23 @@ module.exports = grammar({
...
+    $._raw_multiline_string_middle,
+    $._single_line_string_end,
+    $._multiline_string_end,
     "else",
     "catch",
     "finally",
     "extends",
     "derives",
     "with",
+    $.error_sentinel,
```
jonshea, line 51: The docs suggest having an `error_sentinel` token type. We do not use it, but I found it helpful in debugging.

grammar.js
```diff
@@ -1616,7 +1622,7 @@ module.exports = grammar({
             choice(
               seq(
                 "\\",
-                choice(/[^xu]/, /uu?[0-9a-fA-F]{4}/, /x[0-9a-fA-F]{2}/),
+                choice(/[^xu]/, /[uU]+[0-9a-fA-F]{4}/, /x[0-9a-fA-F]{2}/),
```
jonshea, line 1625: I have a comment noting this elsewhere, but Java weirdly allows any number of `u`s and `U`s at the start of an escape sequence. https://docs.oracle.com/javase/specs/jls/se7/html/jls-3.html

src/scanner.c
```diff
@@ -185,6 +228,24 @@ static inline void debug_indents(Scanner *scanner) {
 
 bool tree_sitter_scala_external_scanner_scan(void *payload, TSLexer *lexer,
                                              const bool *valid_symbols) {
+  #ifdef DEBUG
```
jonshea, line 231: I found this very helpful for debugging.

src/scanner.c
```diff
@@ -388,30 +449,53 @@ bool tree_sitter_scala_external_scanner_scan(void *payload, TSLexer *lexer,
     skip(lexer);
   }
 
-  if (valid_symbols[SIMPLE_STRING] && lexer->lookahead == '"') {
+  if (valid_symbols[SIMPLE_STRING_START] && lexer->lookahead == '"') {
```
jonshea, line 452: Is there a reason we have to scan simple string starts externally instead of just having a normal token, like we do for interpolated strings?

src/scanner.c
```diff
@@ -15,21 +15,53 @@
...
+  "DERIVES",
+  "WITH",
+  "ERROR_SENTINEL"
+};
+
+typedef enum {
+  STRING_MODE_SIMPLE,
+  STRING_MODE_INTERPOLATED,
+  STRING_MODE_RAW,
+} StringMode;
```
eed3si9n, line 63: Could you add comment to document what this mode represents plz?

src/scanner.c
eed3si9n: It looks like this has been around for a long time, and was originally added for interpolated string too https://github.com/tree-sitter/tree-sitter-scala/commit/2dcb6b04b2123a8c9cce2ca0cd43cc40444530e8. It's possible that it was implemented that way back then, but we could move `"` and `"""` into the parser?

eed3si9n (approved):

susliko (approved):

jonshea:
I spot checked some 5 of the 21 files in scala2-compiler that cause the CI failure, and they were all errors caused by a raw string that had a backslash that was not part of a valid escape sequence. For example, `val parseVerbose = raw"\* \S+ \S+ \[([^:]+): .*\] .*".r`. So, I guess we are back at the raw-string discussion from yesterday. Here are my ideas.

1. We could decide not to parse escape sequences in any interpolated strings.
2. We could parse only valid escape sequences in interpolated strings, and treat invalid escape sequences as part of the string instead of treating them like a parse error. 
3. We could go back to treating raw-strings as a special case.

I am in mildly inclined towards 2 or 3, but happy to do whatever the maintainers prefer.

susliko:
> a raw string that had a backslash that was not part of a valid escape sequence

I would suggest making `escape_sequence` rule more relaxed by allowing any character after backslash.

We've encountered a similar situation a few times before. Imo it's okay to have tree-sitter grammar more lenient than actual Scala parser and compiler. I can't see how this approach would hurt main tree-sitter use cases like syntax highlighting, code navigation or repomap building. 

@eed3si9n wdyt?

eed3si9n:
The primary use case for the tree-sitter grammar is for syntax hilighting, so while I'm all for tree-sitter being more relaxed etc to accommodate Scala 2 and 3 etc, in this example of writing a regex inside the raw-interpolater, it might be misleading to highlight anything.

My suggestion would be to restore the raw special case, but add a comment explaining why it needs special treatment.

susliko:
> My suggestion would be to restore the raw special case, but add a comment explaining why it needs special treatment.

Okay, agreed 👍

eed3si9n:
> Do we currently have any way of tracking known syntax that fails to parse? I found a few examples, and I was thinking about making a `test/corpus/failing.txt` to hold them, with the test cases marked with `:skip`. Would that be ok? Or is there something else I should do?

That's a great idea. Failing could mean two things:
1. Pending example (`:skip`), which _should_ parse but currently doesn't
2. Negative example (`:error`), which _should not_ parse and currently doesn't

There are some of them tracked as GitHub issues, but I think it would be cool to have some naming conventions like `test/corpus/literals-pending.txt` and organize them by areas.

PR #474, opened 2025-08-05
feat: applied constructor types

susliko:
Apart from the `tracked` modifier (handled by #473), [Modularity Impovements](https://docs.scala-lang.org/scala3/reference/experimental/modularity.html) introduce one more grammar change: applied constructor types.

```
SimpleType        ::=  SimpleLiteral
                    |  ‘?’ TypeBounds
---                 |  SimpleType1
+++                 |  SimpleType1 {ParArgumentExprs}
```

Example code:
```scala
val c: C(42) /* C { val v: 42 } */ = C(42)
```


grammar.js
```diff
@@ -1366,8 +1369,7 @@ module.exports = grammar({
      * PostfixExpr [Ascription]
      */
     ascription_expression: $ =>
-      prec.dynamic(
```
susliko, line 1369: Lack of fail-fast `left` precedence with conjunction of introduced `applied_constructor_type` could treat a body of the this class
```scala
class B {
  self: Something[A] => 

  case class Hello(a: Int)
}
```
as a giant `ascription_exception` with `Something[A] => \n case class Hello[a]` being a type signature.
image

eed3si9n (approved):

PR #504, opened 2026-04-10
fix: higher precedence for match-clause guards

susliko:
In complex expressions involving pattern-matching cases with 
guards the GLR parser would drop a correct branch during condensing phase. 

Here we add hint GLR with a correct branch via dynamic precedence for the `$._case_pattern`


script/smoke_test.sh
```diff
@@ -3,7 +3,7 @@
 # This is an integration test to generally check the quality of parsing.
 
 SCALA_SCALA_LIBRARY_EXPECTED=100
-SCALA_SCALA_COMPILER_EXPECTED=98
+SCALA_SCALA_COMPILER_EXPECTED=99
```
susliko, line 6: After this change there is the only file in the scala 2 compiler codebase that is parsed with ERROR nodes.

Here's an issue for that: https://github.com/tree-sitter/tree-sitter-scala/issues/505

eed3si9n (approved):
Thanks!

eed3si9n:
I wonder if using `reserved` feature that got added in https://github.com/tree-sitter/tree-sitter/releases/tag/v0.25.0 would fix some of these infix mixups.

susliko:
> I wonder if using `reserved` feature that got added in https://github.com/tree-sitter/tree-sitter/releases/tag/v0.25.0 would fix some of these infix mixups.

I tried it briefly and it looked like it messed up parsing of some files in an unexpected way. Hopefully I'll dig into it

Issue #507, opened 2026-04-11
Scanner does not insert automatic semicolon even if symbolic operator is not followed by operand


susliko:
### Commit of tree-sitter-scala you tested this on

d1722b5c2c1f0818d617bf8a48bcacfa34fd93e0

### A code sample showing the error

```
  final def a = 
    println(1)
    ???
```

### Show the error node

Because of `???` treated as operator identifier here, whole body is treated like a postfix expression
```
(compilation_unit [0, 2] - [4, 0]
  (function_definition [0, 2] - [4, 0]
    (modifiers [0, 2] - [0, 7])
    name: (identifier [0, 12] - [0, 13])
    body: (indented_block [1, 4] - [4, 0]
      (postfix_expression [1, 4] - [2, 7]
        (call_expression [1, 4] - [1, 14]
          function: (identifier [1, 4] - [1, 11])
          arguments: (arguments [1, 11] - [1, 14]
            (integer_literal [1, 12] - [1, 13])))
        (operator_identifier [2, 4] - [2, 7])))))
```

### What do you expect the tree to look like

```
(compilation_unit [0, 2] - [4, 0]
  (function_definition [0, 2] - [4, 0]
    (modifiers [0, 2] - [0, 7])
    name: (identifier [0, 12] - [0, 13])
    body: (indented_block [1, 4] - [4, 0]
      (call_expression [1, 4] - [1, 14]
        function: (identifier [1, 4] - [1, 11])
        arguments: (arguments [1, 11] - [1, 14]
          (integer_literal [1, 12] - [1, 13])))
      (identifier [2, 4] - [2, 5]))))
```

### Where are you experiencing this error?

tree-sitter cli

susliko:
Fixed in #508

PR #508, opened 2026-04-11
fix: suppress automatic semicolon only when infix operator is actually followed by an operand

susliko:
Fixes a bug #507 introduced in #493


eed3si9n (approved):
Thanks!

PR #514, opened 2026-04-17
feat: support dot-syntax for match expressions

### LINKED ISSUE #510: Dot-syntax for match expressions

a06047ee441bd02ca6ebcc0f913737b7c6e178e1

```
this.match
  case self: LocalCap => true
```

### Show the error node

```
(compilation_unit [0, 0] - [3, 0]
  (field_expression [0, 0] - [0, 10]
    value: (identifier [0, 0] - [0, 4])
    field: (identifier [0, 5] - [0, 10]))
  (ERROR [1, 2] - [1, 6])
  (lambda_expression [1, 7] - [1, 29]
    parameters: (identifier [1, 7] - [1, 11])
    parameters: (type_identifier [1, 13] - [1, 21])
    (boolean_literal [1, 25] - [1, 29])))
```

From the [doc](https://docs.scala-lang.org/scala3/reference/changed-features/match-syntax.html):
> match may follow a period:
> ```scala
>if xs.match
>  case Nil => false
>  case _   => true
>then "nonempty"
>else "empty"
>```

```
SimpleExpr   ::=  ...
               |  SimpleExpr ‘.’ MatchClause
```

_No response_

susliko:
Resolves #510


eed3si9n (approved):
Thanks!

PR #516, opened 2026-04-18
feat: allow dot-syntax match expressions in call expression chains

susliko:
We add a dot-syntax match expression into $._simple_expression to allow usage like:
```scala
foo
  .bar()
  .match
    case a =>
    case b =>
  .buz()
```


eed3si9n (approved):
Thanks!

Issue #505, opened 2026-04-10
Misaligned multi-line infix expressions inside a call expression


susliko:
### Commit of tree-sitter-scala you tested this on

717d9eadbcb7bd14dc861840d8402d66759e1d65

### A code sample showing the error

```scala
xs.filter(x =>
     a
  && b
)
```

### Show the error node
```
(compilation_unit [0, 0] - [5, 0]
  (call_expression [0, 0] - [3, 1]
    function: (field_expression [0, 0] - [0, 9]
      value: (identifier [0, 0] - [0, 2])
      field: (identifier [0, 3] - [0, 9]))
    arguments: (arguments [0, 9] - [3, 1]
      (lambda_expression [0, 10] - [2, 6]
        parameters: (identifier [0, 10] - [0, 11])
        (ERROR [1, 5] - [2, 2]
          (indented_block [1, 5] - [2, 2]
            (identifier [1, 5] - [1, 6])))
        (postfix_expression [2, 2] - [2, 6]
          (operator_identifier [2, 2] - [2, 4])
          (identifier [2, 5] - [2, 6]))))))
```
### What do you expect the tree to look like
```
(compilation_unit [0, 0] - [5, 0]
  (call_expression [0, 0] - [3, 1]
    function: (field_expression [0, 0] - [0, 9]
      value: (identifier [0, 0] - [0, 2])
      field: (identifier [0, 3] - [0, 9]))
    arguments: (arguments [0, 9] - [3, 1]
      (lambda_expression [0, 10] - [3, 0]
        parameters: (identifier [0, 10] - [0, 11])
        (indented_block [1, 5] - [3, 0]
          (infix_expression [1, 5] - [2, 9]
            left: (identifier [1, 5] - [1, 6])
            operator: (operator_identifier [2, 5] - [2, 7])
            right: (identifier [2, 8] - [2, 9])))))))
```
### Where are you experiencing this error?

[TypeDiagnostics.scala](https://github.com/scala/scala/blob/5433e535dadb665b25c647b9966ddb7dbeeb24db/src/compiler/scala/tools/nsc/typechecker/TypeDiagnostics.scala) has a few such patterns

susliko:
Fixed by #521

PR #521, opened 2026-04-21
Full coverage of Scala2 compiler

susliko:
This pull requests addresses both grammar issues preventing full coverage of scala 2 compiler codebase. Both issues manifest themselves in [TypeDiagnostics.scala](https://github.com/scala/scala/blob/5433e535dadb665b25c647b9966ddb7dbeeb24db/src/compiler/scala/tools/nsc/typechecker/TypeDiagnostics.scala):
1. https://github.com/tree-sitter/tree-sitter-scala/issues/505
```scala
xs.filter(x =>
     a
  && b
)
```
2. Wildcard in self-type:
```scala
trait C {
  _: Analyzer with StdAttachments =>
  def a = 1
}
```


src/scanner.c
```diff
@@ -247,6 +247,43 @@ static bool scan_word(TSLexer *lexer, const char* const word) {
   return !iswalnum(lexer->lookahead);
 }
 
+// Returns true if the lookahead starts a leading infix operator — a symbolic
+// operator or back-ticked identifier followed by whitespace and then a
+// non-whitespace operand on the same line. Such a line is a continuation of
+// the previous expression, so neither AUTOMATIC_SEMICOLON nor OUTDENT should
+// fire ahead of it. Advances the lexer; the caller must not rely on position.
+static bool is_leading_infix_continuation(TSLexer *lexer) {
```
susliko, line 255: Moved this code from AUTOMATIC_SEMICOLON section into a reusable function in order to suppress OUTDENT token as well

susliko:
@eed3si9n Happy to request a review on this one 🙂

src/scanner.c
```diff
@@ -380,6 +414,17 @@ bool tree_sitter_scala_external_scanner_scan(void *payload, TSLexer *lexer,
     } else {
       scanner->last_column = (int16_t)lexer->get_column(lexer);
     }
+    // Don't close the indented block when the next line starts with a leading
+    // infix operator: that operator continues the previous expression.
+    if (lexer->lookahead != 0 && is_leading_infix_continuation(lexer)) {
+      return false;
+    }
```
eed3si9n: Is it because of the lookahead of leading operator? I feel like this has more to do with the outer parenthesis that normal indent-outdent rules are ignored, like your issue name indicates.

susliko: Yeah, a proper fix here would be to not issue `indent`/`outdent` and `automatic_semicolon` tokens at all when inside parens `(..)` or brackets `[..]`. As far as I understand - this is how the actual Scala parser does it.

I tried an approach of tracking  `({[]})` in external scanner and implementing logic of sort "if we have seen 1 unmatched parenthesis and there were no unmatched brackets or braces, then we are inside `(..)`, do not issue automatic semicolon or indents", but hit into a dead end by breaking a lot of existing cases, so I've parked this approach for now.

susliko: I'll try this approach once again and get back here - maybe I've missed something

susliko: @eed3si9n 
Here is my chain of thought.

1. The following code is invalid in Scala 2 because "automatic semicolons" are not issued for newlines inside parens by its parser, instead they are treated as whitespaces. But its valid in Scala 3.
```scala
xs.map(x =>
   val y = x * 2
   x
)
```
2. Since in tree-sitter we can't distinguish between dialects judging by the contents of the file being parsed, we are forced to have a `$.lambda_expression` that accepts a body with indent/outdent tokens and automatic semicolon tokens.
3. That means that if we want to recognize infix expressions in code like below (valid both in Scala 2 and 3), we need either 
   - allow `$.automatic_semicolon` after the first operand and still somehow tweak issuing `$.unindent`, or 
   - suppress `$.automatic_semicolon` (already works for leading infix operators) and suppress `$.unindent`.
```scala
xs.filter(x =>
     a
  && b
)
```

4. First option does not look viable as everything starts to look as an infix expression - too many things break. Therefore this PR implements the second approach. It's not ideal, but i cannot think about a better way to deal with it.

eed3si9n: ok. Sounds good to me. Thanks for digging into this.

eed3si9n (approved):
Thanks!

PR #527, opened 2026-04-25
feat: allow lazy repeated parameter types

susliko:
Now the grammar defines a repeated param type and a lazy param type as mutually exclusive options.

This PR fixes this:
```scala
case class Foo(a: => A*)
```


eed3si9n (approved):
Thanks!

PR #528, opened 2026-04-25
feat: allow trailing comma in $.indented_cases

susliko:
This fixes cases like this:
```scala
x.apply(
  a match
    case a => a,
)
```


eed3si9n (approved):
Thanks!

PR #529, opened 2026-04-25
feat: allow $.indented_block in $.assignment_expression

susliko:
Fixes code like
```scala
x.copy(views =
  val key = 1
  key + 1
)
```


eed3si9n (approved):
Thanks!

PR #534, opened 2026-04-26
feat: support colon argument with arguments in parenthesis on the same line

susliko:
Adds support for code like this:
```scala
xs.map: (_: Int, a) =>
  a
```

Also
*  removes the dynamic precedence on `$.binding` and 
* inlines `$._simple_type` to resolve ambiguities via meaningful GLR conflicts instead of using `prec.left` for the `$._simple_type`


eed3si9n (approved):
Thanks!

PR #537, opened 2026-05-11
feat: add reserved keywords

### LINKED ISSUE #70: Reserved keywords (val, def) should be rejected as an identifier
## steps

```scala
object A:
  val val = val
```

```bash
tree-sitter parse examples/A.scala
```

## problem

```lisp
(compilation_unit [0, 0] - [2, 0]
  (object_definition [0, 0] - [2, 0]
    body: (template_body [0, 8] - [2, 0]
      (val_definition [1, 2] - [1, 15]
        pattern: (identifier [1, 6] - [1, 9])
        value: (identifier [1, 12] - [1, 15])))))
```

## expectation

`val` should be rejected as an identifier, and it should fail to parse.

## notes

Given that Scala supports infix and postfix expression, combination of `expr identifier expr` would be parsed as an infix expression, and since we don't reject `val`, `def`, etc a small mistake in the grammar ends up spreading widely.

iigaz:
Fixes #70. This PR adds reserved keywords from [Scala 2](https://www.scala-lang.org/files/archive/spec/2.13/01-lexical-syntax.html) and [Scala 3](https://docs.scala-lang.org/scala3/reference/syntax.html) (except "⇒" (`\u21D2`) and "←" (`\u2190`), reserved in Scala 2 but not used in this repo anyway, and `forSome`, which is not implemented here yet), using tree-sitter API added in https://github.com/tree-sitter/tree-sitter/pull/3896.

---

While I was testing it I encountered an issue which is not directly connected to reserved keywords, but is more visible with them. The issue is that `this` and `super` are treated as valid identifiers in places where they should normally be not allowed (e.g. `val this = ???`), which breaks their `reserved` status (moreover, they aren't really identifiers per spec). Moving them out of `identifiers` is a lot of changes (I initially [tried to do it](https://github.com/iigaz/tree-sitter-scala/pull/1/files) but gave up after several corner-cases), and IMO out of scope of this PR. Please let me know if you think it should be addressed here.


src/scanner.c
```diff
@@ -520,6 +516,10 @@ bool tree_sitter_scala_external_scanner_scan(void *payload, TSLexer *lexer,
       }
     }
 
+    if (valid_symbols[ELSE]) {
+      return !scan_word(lexer, "else");
+    }
+
```
iigaz: For some reason, adding `else` in `reserved` broke "Try expressions" on lines where `catch` and `finally` are just after indentation, and changing the order of scanning here fixed it.

susliko:
@iigaz Thanks for giving it a shot!

As far as i can tell, this is what happens here:
* `reserved` feature somehow tempers with external tokens (i.e. tokens, produced by scanner.c). This might be a bug or just an undocumented behavior of tree-sitter itself.
* we have `else`, `finally`, etc. tokens defined as external in the first place to avoid emitting AUTOMATIC_SEMICOLON token in the middle of expressions like
 ```scala
try a()
finally c()
```

So for now you could narrow down the 'reserved' list to only those, which are not emitted by the external scanner, i.e. leave `else`, `catch`, `finally`, `extends`, `if`, `derives` as valid identifiers for now.

A proper fix, in my opinion, would be to include optional `$._automatic_semicolon` in if/try expressions instead of trying to suppress its emission via an external scanner. I gave it a quick try and as a side effect it reduces parser size from 34M to 14M

**Comment by iigaz** [comment]
Thanks for taking a look! I removed operations from reserved words (they're not as important to reserve as regular words IMO but break a lot of tests), and commented out problematic keywords that I think should be reserved, but currently can't because they break some of the tests (most of which really were broken *before* these words became reserved):

- `catch` breaks in cases described in #540.
- `if` breaks on incorrect else-if parsing as a consequence of #541.
- `then` also breaks in #541.
- ~~`new`~~ [also breaks as a consequence](https://github.com/scala/scala3/blob/65a6143a129929df825e3693209ead96de5a4e49/compiler/src/dotty/tools/dotc/typer/ProtoTypes.scala#L421) of #541.
- `with` breaks in #538.
- ~~`finally`~~ [breaks as a consequence](https://github.com/scala/scala3/blob/65a6143a129929df825e3693209ead96de5a4e49/compiler/src/dotty/tools/dotc/core/Types.scala#L5325) of #542.
- `else` breaks tests due to automatic semicolon (thanks!).

I will try to fix some of the issues here, but this PR might have to wait until that.

Edit: I discovered these issues on latest Dotty, and just noticed that you are running smokes on 3.5.2 so I uncommented some keywords not directly related to issues above, this should keep parsing degradation on CI within acceptable limits.

File: grammar.js [hunk 2]
```diff
@@ -68,7 +68,7 @@ module.exports = grammar({
   // These names can be used in the prec functions to define precedence relative only to other names in the array, rather than globally.
   precedences: $ => [
     ["mod", "soft_id"],
-    ["end", "soft_id"],
+    ["end", "this_id", "soft_id"],
```
susliko: Is this change necessary? I.e. in which case a conflict arises between end-marker, `this`, and soft identifiers?

grammar.js
```diff
@@ -160,6 +160,55 @@ module.exports = grammar({
...
+      'finally',
+      'for',
+      // 'forSome', // Scala 2, not implemented
+      'given', // Scala 3
+      // 'if',
+      'implicit',
+      'import',
+      'lazy',
+      'macro', // Scala 2
+      token.immediate('match'),
```
susliko, line 190: Are you sure that `token.immediate` does something meaningful here?

It's a way to describe tokens that go immediately after previous tokens, without whitespaces in the middle.
[doc](https://tree-sitter.github.io/tree-sitter/creating-parsers/2-the-grammar-dsl.html?highlight=token.immediate#the-grammar-dsl)

susliko:
> I will try to fix some of the issues here, but this PR might have to wait until that.

I think it's totally fine to merge this with only a subset of reserved keywords as it already improves the error recovery

grammar.js
```diff
@@ -160,6 +160,55 @@ module.exports = grammar({
 
   word: $ => $._alpha_identifier,
 
+  reserved: {
+    global: $ => [
```
susliko, line 165: Did you try adding symbolic keywords to this list?
```
:         =         <-        =>        <:        >:        #
@         =>>       ?=>
```

`=>` would be especially useful for a proper error recovery as it's widely used in lambda expressions and function types.

grammar.js
iigaz: Yes, I'm sure. Doing just "match" breaks a test case with dot-syntax for match expressions, and its definition uses `token.immediate`. I too found this quite bizarre, but it both passes the test and reserves "match".

grammar.js
iigaz: Yes, actually, they only got removed in the latest commit because they broke way too many smoke tests. Come to think of it, this also can be used as an opportunity to find (and hopefully fix) more parsing bugs, so I'll add them back in with `<%` from scala 2 (spec also mentions `_` as reserved, but given its close connection to identifiers I don't think it's necessary to put it here).

grammar.js
iigaz: There's a conflict only between `end` and `this_id`. Likely I just took a path of least resistance by putting `this_id` here, I'll take it out to a separate list.

grammar.js
```diff
@@ -160,6 +161,69 @@ module.exports = grammar({
...
+      "=",
+      "<-",
+      // "=>",
+      // "<:",
+      "<%", // Scala 2
+      // ">:",
+      "#",
+      "@",
+      "=>>",
+      "?=>",
```
iigaz, line 223: `=>`, `<:`, `>:` break some smoke tests in Scala 2.

`#`, `?=>` break some smoke tests in Scala 3 (but not enough to fail the CI).

I haven't dug into them yet, I will file bug reports after I do.

UPD:
- `=>` (and `?=>`): #543.
- `<:` and `>:` are treated as operators in infix types (e.g. in `T[_ >: B]`), and I can't really tell if it's intentional or not.
- `#` breaks on hash-comments (for example, on [this line](https://github.com/scala/scala3/blob/65a6143a129929df825e3693209ead96de5a4e49/compiler/test-resources/scripting/hashBang.scala#L2)). This is not valid Scala afaik, so I don't think this is a bug.

eed3si9n (approved):
Thanks for the contribution!

PR #544, opened 2026-05-14
fix: applied_constructor_type and early defs

### LINKED ISSUE #538: Instance expression with `with` is parsed as infix expression

a68000002745b94eec61cef741efe7cede4ff465

```scala
new Stepper.UnboxingIntStepper(st) with EfficientSplit
```

### Show the error node

```
  (infix_expression [0, 0] - [0, 54]
    left: (instance_expression [0, 0] - [0, 34]
      (stable_type_identifier [0, 4] - [0, 30]
        (identifier [0, 4] - [0, 11])
        (type_identifier [0, 12] - [0, 30]))
      arguments: (arguments [0, 30] - [0, 34]
        (identifier [0, 31] - [0, 33])))
    operator: (identifier [0, 35] - [0, 39])
    right: (identifier [0, 40] - [0, 54])))
```

_No response_

_No response_

iigaz:
Fixes #538, and adds support for early definitions.

Early definitions here are done via braced_template_body, which is not entirely correct per se because early defs do not contain statements and are esentially just `val` and `var` definitions, but I had some trouble making "letter-perfect" early defs, and this is the simplest way to correctly parse correct code.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -774,7 +774,6 @@ module.exports = grammar({
           // This adds _simple_type, but not the above intentionally.
           seq($._simple_type, field("arguments", $.arguments)),
           seq($._annotated_type, field("arguments", $.arguments)),
-          seq($.compound_type, field("arguments", $.arguments)),
         ),
       ),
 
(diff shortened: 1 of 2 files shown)


grammar.js
```diff
@@ -774,7 +774,6 @@ module.exports = grammar({
           // This adds _simple_type, but not the above intentionally.
           seq($._simple_type, field("arguments", $.arguments)),
           seq($._annotated_type, field("arguments", $.arguments)),
-          seq($.compound_type, field("arguments", $.arguments)),
```
iigaz, line 777: Removed because `arguments` are inside the `compound_type`, not after it

eed3si9n (approved):
Thanks!

PR #547, opened 2026-07-20
feat: reduce real-world parse failures across 34 OSSs

exoego:
Parses 14,336 / 14,446 (99.24%) of real-world Scala files

Measured against akka, almond, better-files, Binding.scala, caliban, cats-collections, circe, eff, http4s, iron, kamon, macwire, monocle, parquet4s, play-json, playframework, refined, Scala 2, Scala 3 (except the experimental capture checking), scala-logging, scala-xml, scalacache, scalacheck, scalamock, scalikejdbc, scopt, shapeless, skunk, slick, spire, squants, sttp, tapir, zio.

The below issues are closed by this PR:
Closes #264
Closes #329
Closes #358 
Closes #465 
Closes #539 
Closes #540
Closes #541
Closes #542

Additionally, two optimizations were added since Parser.c neared its 33 MiB ceiling.
1. Inlining eight hidden rules cut large-state count 4495→2875 (−2.2 MiB).
2. Moving block comments to the external scanner removed a slash-star `/*` column from every parse state (−460 KB).

(Parser grew up a bit for Scala2/3 repo after these two optimization)

Net: 29.6 MiB (3.4 MiB headroom), ~30% faster parsing, tree shape unchanged.

PR #549, opened 2026-07-21
fix(scala3): allow an indented block as the body of `catch case`

### LINKED ISSUE #542: `catch case` should accept a block instead of an expression

a680000

As seen in https://github.com/scala/scala3/blob/65a6143a129929df825e3693209ead96de5a4e49/compiler/src/dotty/tools/dotc/core/Types.scala#L5325

```scala
try
  ???
catch case ex: Throwable =>
  ???
  ???
```

### Show the error node

```
(compilation_unit [0, 0] - [5, 0]
  (try_expression [0, 0] - [3, 5]
    body: (indented_block [1, 2] - [2, 0]
      (operator_identifier [1, 2] - [1, 5]))
    (catch_clause [2, 0] - [3, 5]
      pattern: (typed_pattern [2, 11] - [2, 24]
        pattern: (identifier [2, 11] - [2, 13])
        type: (type_identifier [2, 15] - [2, 24]))
      body: (operator_identifier [3, 2] - [3, 5])))
  (operator_identifier [4, 2] - [4, 5]))
```

```
(compilation_unit [0, 0] - [5, 0]
  (try_expression [0, 0] - [3, 5]
    body: (indented_block [1, 2] - [2, 0]
      (operator_identifier [1, 2] - [1, 5]))
    (catch_clause [2, 0] - [3, 5]
      pattern: (typed_pattern [2, 11] - [2, 24]
        pattern: (identifier [2, 11] - [2, 13])
        type: (type_identifier [2, 15] - [2, 24]))
      body: (operator_identifier [3, 2] - [3, 5])
      body: (operator_identifier [4, 2] - [4, 5]))))
```

tree-sitter CLI

---

P.S. Unfortunately I don't see it in the specification, but this code compiles and prints 123 on Scala 3.1.3, 3.4.3 and 3.8.3:

```scala
@main
def main =
  try
    ???
  catch case ex: Throwable =>
    val test = 123
    println(test)
```

Which means:
1. definitions inside `catch case` are valid;
2. `catch case` may contain several lines.

And based on this, I conclude that `$._expr_case_clause` might accept `$._block` in `"body"` (like regular `$.case_clause`) instead of `$.expression`. But then again, this [contradicts the specification](https://www.scala-lang.org/files/archive/spec/3.4/08-pattern-matching.html#pattern-matching-expressions) which clearly says `Expr`...

exoego:
- Spin-out from #547
- Closes #542

A brace-less `catch case pat =>` whose body spans multiple indented lines was cut off after the first statement; the remaining statements leaked out of the try expression (issue #542, seen in real code such as [dotty's Types.scala)](https://github.com/scala/scala3/blob/03f6bdd2818b237d356828675f90c920158246e5/compiler/src/dotty/tools/dotc/core/Types.scala#L5316-L5318). Accept `$.indented_block` as the clause body: the block's OUTDENT closes it, so following statements or a `finally` clause are not swallowed.

The expected tree of the existing multi-line catch-case test now wraps the body in `indented_block`.
A legitimate shape change that comes with accepting a block there.

### DIFF
--- a/test/corpus/expressions.txt
+++ b/test/corpus/expressions.txt
@@ -783,7 +783,8 @@ def main(): Unit =
           (typed_pattern
             (identifier)
             (type_identifier))
-          (identifier))))))
+          (indented_block
+            (identifier)))))))
 
 ================================================================================
 Match expressions
@@ -2161,3 +2162,41 @@ bar(a, xs: _*)
       (identifier)
       (vararg
         (identifier)))))
+
+================================================================================
+Catch case with a multi-statement indented body
+================================
+
+// https://github.com/tree-sitter/tree-sitter-scala/issues/542
+try
+  ???
+catch case ex: Throwable =>
+  val test = 123
+  println(test)
+finally
+  cleanup()
+
+---
+
+(compilation_unit
+  (comment)
+  (try_expression
+    (indented_block
+      (operator_identifier))
+    (catch_clause
+      (typed_pattern
+        (identifier)
+        (type_identifier))
+      (indented_block
+        (val_definition
+          (identifier)
+          (integer_literal))
+        (call_expression
+          (identifier)
+          (arguments
+            (identifier)))))
+    (finally_clause
+      (indented_block
+        (call_expression
+          (identifier)
+          (arguments))))))
--- a/grammar.js
+++ b/grammar.js
@@ -1488,7 +1488,13 @@ module.exports = grammar({
       ),
 
     _expr_case_clause: $ =>
-      prec.left(seq("case", $._case_pattern, field("body", $.expression))),
+      prec.left(
+        seq(
+          "case",
+          $._case_pattern,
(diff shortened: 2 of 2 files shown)


test/corpus/expressions.txt
```diff
@@ -2161,3 +2162,41 @@ bar(a, xs: _*)
...
+================================================================================
+Catch case with a multi-statement indented body
+================================
+
+// https://github.com/tree-sitter/tree-sitter-scala/issues/542
+try
+  ???
+catch case ex: Throwable =>
+  val test = 123
+  println(test)
```
exoego, line 2175: [Real-world example in scala3]( https://github.com/scala/scala3/blob/03f6bdd2818b237d356828675f90c920158246e5/compiler/src/dotty/tools/dotc/core/Types.scala#L5316-L5318)

susliko:
Thank you for the contribution!

PR #550, opened 2026-07-21
fix: accept braced-body members indented shallower than their siblings

exoego:
- Spin-out from #547 
- Corresponding issue: None. [Real-world example in shapeless](https://github.com/milessabin/shapeless/blob/99efca64f4f5fe27c7b1c3391403fed1ba3f3320/core/shared/src/main/scala/shapeless/ops/hlists.scala#L1065-L1069)

Inside a braced template body, the scanner opens an indent region at the first member; a later member indented shallower than that (legal, scalac ignores indentation inside braces) popped the region before the closing brace and the remaining members fell out of the body as parse errors.

Let `_braced_template_body2` continue with a second statement block after the early OUTDENT, so the region close no longer ends the braced body.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -550,6 +550,10 @@ module.exports = grammar({
         ),
         optional($._block),
         $._outdent,
+        // A member indented shallower than its siblings pops the indent
+        // region before the closing brace; the remaining members are still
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Thanks!

PR #551, opened 2026-07-21
fix: self types with empty bodies, empty-bodied lambdas, enum self types

exoego:
- Corresponding issue: None.

Four related gaps around SelfType  and empty lambda bodies:

1. `foo { x => }` — a block lambda with an empty body was a parse error; the body block is now optional. This grammar change is included here  because the empty-body self-type fix below builds on it.
    Seen in [akka](https://github.com/akka/akka-core/blob/e75e809380ee16a18b12007d69bb46c30b9230e7/akka-stream-tests-tck/src/test/scala/akka/stream/tck/ForeachSinkSubscriberTest.scala#L15-L16)/[pekko](https://github.com/apache/pekko/blob/4fe1f4a2d98c43469b196123f032daf1da9e0410/stream-tests-tck/src/test/scala/org/apache/pekko/stream/tck/ForeachSinkSubscriberTest.scala#L25-L26).
2. `trait A { self: B => }` — a braced template body that is only a self type had no production. The empty-bodied-lambda reading must complete without error recovery so that the GLR fork reaches condense, where self_type's dynamic precedence wins; with the lambda branch still erroring, the recovery version shadowed the self-type reading.
   Seen in [circe](https://github.com/circe/circe/blob/e884a42003270af4738e32442768de27c266452f/modules/tests/js/src/test/scala/io/circe/FloatJsonTests.scala#L25) and [scala-xml](https://github.com/scala/scala-xml/blob/a4c4e52e88ea92fb019650208984170942407db4/shared/src/main/scala-2.12/scala/xml/ScalaVersionSpecific.scala#L38-L44).
3. `trait A:` + `self: B =>` with no members after it — the same gap for colon template bodies. 
     Seen in [dotty](https://github.com/scala/scala3/blob/03f6bdd2818b237d356828675f90c920158246e5/compiler/src/dotty/tools/dotc/reporting/Diagnostic.scala#L44-L46).
4. [enum bodies did not accept a self type ](https://github.com/scala/scala3/blob/03f6bdd2818b237d356828675f90c920158246e5/compiler/src/dotty/tools/dotc/semanticdb/Scala3.scala#L161-L162)at all, and [enum cases did not accept modifiers](https://github.com/scala/scala3/blob/03f6bdd2818b237d356828675f90c920158246e5/compiler/src/dotty/tools/io/FileExtension.scala#L16).
     Seen in dotty.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -280,10 +280,19 @@ module.exports = grammar({
 
     enum_body: $ =>
       choice(
-        prec.left(PREC.control, seq(":", $._indent, $._enum_block, $._outdent)),
+        prec.left(
+          PREC.control,
+          seq(
+            ":",
+            $._indent,
+            optional($.self_type),
+            $._enum_block,
+            $._outdent,
+          ),
+        ),
         seq(
           "{",
-          // TODO: self type
+          optional($.self_type),
           optional($._enum_block),
           "}",
         ),
@@ -292,6 +301,8 @@ module.exports = grammar({
     enum_case_definitions: $ =>
       seq(
         repeat($.annotation),
+        // e.g. `private case External(...) extends ...`
+        optional($.modifiers),
         "case",
         choice(commaSep1($.simple_enum_case), $.full_enum_case),
       ),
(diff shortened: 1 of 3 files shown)


eed3si9n (approved):
Thanks!

PR #555, opened 2026-07-22
fix: keep soft keyword `inline` usable as an identifier in expressions

exoego:
- Corresponding issue: None.
 
`inline` is the only soft modifier that is also valid at expression start (`inline if`, `inline match`), so the static "mod" > "soft_id" precedence made the parser commit to inline_modifier and break plain identifier uses such as `!inline || x`.
Seen in[ playframework Results.scala](https://github.com/playframework/playframework/blob/cb6b3b9938746723d0bbea7f1c7ac03cd4361f4c/core/play/src/main/scala/play/api/mvc/Results.scala#L556-L557)

The token is now reduced under a GLR fork declared as a conflict between inline_modifier and _soft_identifier; the continuation picks the surviving reading, and a dynamic precedence on inline_modifier prefers the modifier when both readings complete (`inline if (c) a else b`).

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -78,6 +78,7 @@ module.exports = grammar({
 
   conflicts: $ => [
     [$.tuple_type, $.parameter_types],
+    [$.inline_modifier, $._soft_identifier],
     [$.binding, $._simple_expression],
     [$.binding, $._type_identifier],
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Thanks!

PR #556, opened 2026-07-22
feat: method values (`f _`) as simple expressions

exoego:
- Corresponding issue: None.

`expr _` had no rule of its own. The underscore lexed as an identifier and the whole expression parsed as a postfix expression. A postfix expression cannot be an infix operand, so `(iso.reverseGet _ compose iso.get)(s)` broke. A new method_value simple expression covers it. Where it is valid the underscore now lexes as the wildcard keyword token, so the old postfix reading simply disappears.

Seen in [monocle IsoLaws.scala](https://github.com/optics-dev/Monocle/blob/3aa940b81c5ff3bfd40ec438ae0580b5b59c8bfe/core/shared/src/main/scala/monocle/law/IsoLaws.scala#L12-L13).

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -1398,8 +1398,16 @@ module.exports = grammar({
         $.field_expression,
         $.generic_function,
         $.call_expression,
+        $.method_value,
         alias($._dot_match_expression, $.match_expression),
       ),
+
+    /**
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Thanks!

PR #557, opened 2026-07-22
fix: restrict end marker names to alphanumeric identifiers

exoego:
- Corresponding issue: None.

The end-marker rule accepted any identifier as the marker name, including operator identifiers. The precedence of "end" over "soft_id" then committed `end` plus an operator to the end-marker reading. `${end - start}` inside an interpolation parsed `end -` as a marker and broke. An end marker for a symbolic definition such as `end ::` is vanishingly rare compared to `end` as a value name followed by an infix operator. dotty also recognizes a marker only when the line contains nothing else.

Seen in [macwire TimingInterceptor.scala](https://github.com/softwaremill/macwire/blob/848ac13ef73a174009b02abf89a51274cdeb42aa/examples/scalatra/src/main/scala/com/softwaremill/macwire/examples/scalatra/util/TimingInterceptor.scala#L14).

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -619,7 +619,13 @@ module.exports = grammar({
             "given",
             "extension",
             "val",
-            alias($._identifier, "_end_ident"),
+            // Only alphanumeric (or back-quoted) names: allowing operator
+            // identifiers here would swallow `end` used as a plain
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Thanks!

PR #558, opened 2026-07-22
feat: Scala 2 script header `#!...!#`

exoego:
- Corresponding issue: None.

A Scala 2 script may open with a shell preamble that runs until a closing `!#` line. Only the first line was recognized, so the rest of the preamble parsed as Scala and produced errors. The shebang token now optionally consumes the remaining preamble lines and the closing `!#`.

Seen in [akka scripts/authors.scala](https://github.com/akka/akka-core/blob/e75e809380ee16a18b12007d69bb46c30b9230e7/scripts/authors.scala#L1-L3).


eed3si9n (approved):
Thanks!

PR #559, opened 2026-07-22
feat(scala2): `forSome` existential types

exoego:
- Corresponding issue: None.

Add existential_type for `P[T] forSome { type T }`. Tree-sitter keyword extraction keeps `forSome` usable as a plain identifier elsewhere. Highlight `forSome` as a keyword.

Seen in [shapeless package.scala.](https://github.com/milessabin/shapeless/blob/99efca64f4f5fe27c7b1c3391403fed1ba3f3320/core/shared/src/main/scala/shapeless/package.scala#L60)

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -1082,6 +1082,13 @@ module.exports = grammar({
         $.literal_type,
         $._structural_type,
         $.type_lambda,
+        $.existential_type,
+      ),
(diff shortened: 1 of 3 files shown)


eed3si9n (approved):
Thanks!

PR #561, opened 2026-07-22
fix(scala2): accept Unicode `⇒` as `=>` and `←` as `<-`

exoego:
- Corresponding issue: None.

Scala 2 (SLS 1.1) treats `⇒` as an alternate spelling of the reserved `=>` token. `←` is likewise accepted for `<-` in for-comprehension generators. Real-world Scala 2 code still uses both.

`⇒` is added through a shared fatArrow() helper at every internal `=>` site, so the tree shape and node names are unchanged and no conflicts are added.

`⇒` cannot collide with user-defined operators because the single-opchar class of operator_identifier excludes U+21D2. `?=>` and `=>>` are Scala-3-only and have no Unicode spellings.

Seen in [kamon DatadogAPIReporter.scala](https://github.com/kamon-io/Kamon/blob/a376f39873921a1b2def82942e06950756522307/reporters/kamon-datadog/src/main/scala/kamon/datadog/DatadogAPIReporter.scala#L115) (`⇒`) and [akka VotingServiceSpec.scala](https://github.com/akka/akka-core/blob/e75e809380ee16a18b12007d69bb46c30b9230e7/samples/akka-sample-distributed-data-java/src/multi-jvm/scala/sample/distributeddata/VotingServiceSpec.scala#L70) (`←`).

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -20,6 +20,14 @@ const PREC = {
   binding: 10,
 };
 
+// `⇒` (U+21D2) is the alternate Scala 2 spelling of `=>` (SLS 1.1, dropped in
+// Scala 3). It is not lexable as an operator_identifier (the single-opchar
+// class excludes it), so the extra token cannot collide with user operators.
+const fatArrow = () => choice("=>", alias("⇒", "=>"));
+
+// `=>` or the context-function arrow `?=>`.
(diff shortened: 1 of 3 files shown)


eed3si9n (approved):
Thanks!

PR #563, opened 2026-07-23
feat: quoted keyword literals `'null` / `'true` / `'false`

exoego:
- Corresponding issue: None.

Scala 3 allows quoting keyword literals in macro code. `quote_expression` now accepts `null_literal` and `boolean_literal` after `'`. `'this` already parses via identifier, so no change is needed for it.

Seen in [almond Logger.scala.](https://github.com/almond-sh/almond/blob/3f7b0a6cd5f105ee54165244149c23620802e55c/modules/shared/logger/src/main/scala-3/almond/logger/Logger.scala#L13)

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -1809,7 +1809,13 @@ module.exports = grammar({
         PREC.macro,
         seq(
           "'",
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Thanks!

PR #564, opened 2026-07-23
fix: allow empty statements before a statement in blocks

exoego:
- Corresponding issue: None.

scalac accepts extra empty statements around any statement separator. `_block` now uses `_semis` as the separator, a `;` or newline followed by extra `;`. Leading semicolons like `;{ ... }` and `;def` get a dedicated choice branch. The semicolon-only block `{ ;; }` stays valid through that same branch.

Seen in slick [doc/code Connection.scala](https://github.com/slick/slick/blob/6d24b1511e54a1b024e397f4474e39afbae05cba/doc/code/Connection.scala#L54), [OrmToSlick.scala ](https://github.com/slick/slick/blob/6d24b1511e54a1b024e397f4474e39afbae05cba/doc/code/OrmToSlick.scala#L69)and [DBIOCombinators.scala.](https://github.com/slick/slick/blob/6d24b1511e54a1b024e397f4474e39afbae05cba/doc/code/DBIOCombinators.scala#L13)

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -146,7 +146,7 @@ module.exports = grammar({
     [$.binding, $._simple_expression, $._type_identifier],
     [$.class_parameter, $._type_identifier],
     // '{'  _single_lambda_param  '=>'  expression  •  '}'  …
-    [$._block, $._indentable_expression],
+    [$._block_statements, $._indentable_expression],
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Thanks!

PR #565, opened 2026-07-23
feat: read `|` after a definition type annotation as a union type

exoego:
- Corresponding issue: None.

`val a: AnyRef | Null = null` misparsed the `|` as an alternative_pattern. Declarations like `var b: Link[?] | Null` were errors. [SLS 4.1 ](https://scala-lang.org/files/archive/spec/3.4/04-basic-definitions.html#value-definitions)restricts the pattern of a val/var definition to Pattern2. The new `_definition_pattern` rule excludes top-level alternative and typed patterns. Infix pattern operands are restricted too because the left recursion would re-admit them.

The two `tuple_pattern` conflicts become unnecessary with this restriction and are removed. The generated parser is byte-identical with or without them.

Seen in [scala3 StreamExtensions.scala](https://github.com/scala/scala3/blob/edb79837372a79f8060bfff0dee641ced94d085c/library/src/scala/collection/convert/StreamExtensions.scala#L534).

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -107,6 +107,8 @@ module.exports = grammar({
     [$._full_enum_def],
     // _start_val  identifier  ','  identifier  •  ':'  …
     [$.identifiers, $.val_declaration],
+    [$.val_declaration, $._definition_pattern],
+    [$.var_declaration, $._definition_pattern],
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
> ok, dotty/compiler/: 94.84%, expected at least 93%

🎉

PR #567, opened 2026-07-24
fix: accept `do` and bare indented bodies after parenthesized for enumerators

exoego:
- Corresponding issue: None.

`for (x <- xs) do body` and a bare indented body after `for (...)` were errors. Adding a `do` branch there while do_while stays inside `$.expression` would make `do` ambiguous after `for (...)` and grow parser.c by about 7MiB. Instead `do_while_expression` moves out of `$.expression`. The new `statementExpression` helper re-admits it in statement and RHS positions. Assignment right-hand sides and `return` keep accepting do-while through the helper.

Seen in scala3 [ParsedLogicalPackage.scala](https://github.com/scala/scala3/blob/4dae25087df0ba25b8f1b05a4337fe9d73f79aa2/compiler/src/dotty/tools/dotc/interactive/ParsedLogicalPackage.scala#L97), [Typer.scala ](https://github.com/scala/scala3/blob/4dae25087df0ba25b8f1b05a4337fe9d73f79aa2/compiler/src/dotty/tools/dotc/typer/Typer.scala#L3068)and [Run.scala](https://github.com/scala/scala3/blob/4dae25087df0ba25b8f1b05a4337fe9d73f79aa2/compiler/src/dotty/tools/dotc/Run.scala#L417-L418)

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -28,6 +28,11 @@ const fatArrow = () => choice("=>", alias("⇒", "=>"));
 // `=>` or the context-function arrow `?=>`.
 const anyArrow = () => choice(fatArrow(), "?=>");
 
+// An expression as admitted in statement and RHS positions. do_while lives
+// outside $.expression: after `for (...)` a `do` must introduce the for's own
+// body, and a do-while fork there doubles the paren-for automaton (~+7MiB).
+const statementExpression = ($) => choice($.expression, $.do_while_expression);
+
 module.exports = grammar({
   name: "scala",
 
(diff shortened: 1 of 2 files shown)


test/corpus/definitions.txt
```diff
@@ -2328,3 +2328,82 @@ object A {
             (identifier)
             (arguments
               (identifier))))))))
+
+================================================================================
+Do-while in assignment and return positions
+================================================================================
```
eed3si9n: I think this should be in expressions.txt, since according the language spec this used to be called do loop expression - https://scala-lang.org/files/archive/spec/2.11/06-expressions.html#do-loop-expressions

exoego: I see. Moved.

eed3si9n (approved):
Thanks!

PR #568, opened 2026-07-24
fix: ascription to function types and ascribed match expressions

exoego:
- Corresponding issue: None.

`(f: Int => Int)` and `(1 second: Duration)` and `({ ... }: Any => Int)` were errors. `ascription_expression` gains an optional arrow tail after the ascribed type. A bare identifier parameter shares its `identifier '=>'` token path with the colon_argument lambda start. `match { ... }: @unchecked` works by admitting match_expression on the left. The infix colon-argument branch is removed.
It statically shadowed postfix ascriptions and postfix_expression plus call_expression already covers `a op:` bodies. The SIP-44 expected tree for a backquoted operator taking a colon argument changes accordingly.

Seen in akka [EventBus.scala ](https://github.com/akka/akka-core/blob/e75e809380ee16a18b12007d69bb46c30b9230e7/akka-actor/src/main/scala/akka/event/EventBus.scala#L299-L300)and [DurationSpec.scala](https://github.com/akka/akka-core/blob/e75e809380ee16a18b12007d69bb46c30b9230e7/akka-actor-tests/src/test/scala/akka/util/DurationSpec.scala#L90-L94), scalamock [OverloadedMethodsTest.scala](https://github.com/scalamock/scalamock/blob/4562f63e5b8ec8aecaa720652648e5157bb09fc6/core/shared/src/test/scala/com/paulbutcher/test/mock/OverloadedMethodsTest.scala#L51) and [NewApiSpec.scala](https://github.com/scalamock/scalamock/blob/4562f63e5b8ec8aecaa720652648e5157bb09fc6/core/shared/src/test/scala-3/newapi/NewApiSpec.scala#L566), and scala3 [Scala2Unpickler.scala](https://github.com/scala/scala3/blob/4dae25087df0ba25b8f1b05a4337fe9d73f79aa2/compiler/src/dotty/tools/dotc/core/unpickleScala2/Scala2Unpickler.scala#L83-L86).

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -28,6 +28,9 @@ const fatArrow = () => choice("=>", alias("⇒", "=>"));
 // `=>` or the context-function arrow `?=>`.
 const anyArrow = () => choice(fatArrow(), "?=>");
 
+const ascriptionArrowTail = ($) =>
+  seq(anyArrow(), field("return_type", $._param_type));
+
 // An expression as admitted in statement and RHS positions. do_while lives
 // outside $.expression: after `for (...)` a `do` must introduce the for's own
 // body, and a do-while fork there doubles the paren-for automaton (~+7MiB).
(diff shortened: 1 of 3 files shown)


eed3si9n (approved):
Thanks!

PR #569, opened 2026-07-24
fix: capture patterns and free escapes inside interpolated strings

exoego:
- Corresponding issue: None.

Quasiquote patterns like `case q"${name @ Ident(_)}" =>` were errors. `interpolation` gains a capture_pattern branch with a dynamic penalty. The penalty keeps the pattern grammar out of expression interpolations. The branch also covers val definition patterns like `val s"${name @ _}/" = x`. `escape_sequence` also accepts any other escaped character like `\ `.
scalac accepts those at the lexer level and leaves validity to the interpolator.

Seen in zio [Zio2Upgrade.scala](https://github.com/zio/zio/blob/13a1ac1af36cfdb0bd7231c073f94ea0f239c0ea/scalafix/rules/src/main/scala/fix/Zio2Upgrade.scala#L646), playframework ScalaSirdRouter.scala, play-json [JsMacroImpl.scala ](https://github.com/playframework/playframework/blob/0fbcf712635e1f3fa84d6a6e97521efc39e13f84/documentation/manual/working/scalaGuide/advanced/routing/code/ScalaSirdRouter.scala#L189-L190)and scala3 [Scala3.scala](https://github.com/scala/scala3/blob/4dae25087df0ba25b8f1b05a4337fe9d73f79aa2/compiler/src/dotty/tools/dotc/semanticdb/Scala3.scala#L187-L188).

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -2059,7 +2059,18 @@ module.exports = grammar({
       alias($._interpolation_identifier, $.identifier),
 
     interpolation: $ =>
-      seq("$", choice($._aliased_interpolation_identifier, $.block)),
+      seq(
+        "$",
+        choice(
+          $._aliased_interpolation_identifier,
+          $.block,
+          // In pattern position an interpolation may hold a capture pattern,
+          // e.g. `case q"${name @ Ident(_)}" =>`. Restricted to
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Thanks!

PR #572, opened 2026-07-24
feat: Unicode math symbols as single-character type operators

exoego:
- Corresponding issue: None.

Type operators like `∧` and `∨` were errors.
SLS 1.1 admits [Unicode math symbols (Sm) as opchars.](https://github.com/scala/scala3/blob/4dae25087df0ba25b8f1b05a4337fe9d73f79aa2/docs/_docs/internals/syntax.md#L43) The single-character opchar class gains the Mathematical Operators block `U+2200-U+22FF`. The full `Sm` category stays excluded because it contains `=`.

Seen in shapeless [package.scala](https://github.com/milessabin/shapeless/blob/99efca64f4f5fe27c7b1c3391403fed1ba3f3320/core/shared/src/main/scala/shapeless/package.scala#L30-L36).


eed3si9n (approved):
Thanks!

PR #573, opened 2026-07-24
fix: soft keyword `extension` and Scala 3 keywords in import paths

exoego:
- Corresponding issue: None.
 
`extension.journalFor(null)` and `import io.circe.export.Exported` were errors. `extension` joins the soft identifier list with a GLR conflict against extension_definition. Import and export paths accept `enum` and `export` segments through a new `_namespace_path_segment` rule. Those are valid identifiers in Scala 2 sources and appear as package names. The highlight query now marks `extension` as a keyword only in extension definitions.

Seen in akka [JournalSpec.scala](https://github.com/akka/akka-core/blob/e75e809380ee16a18b12007d69bb46c30b9230e7/akka-persistence-tck/src/main/scala/akka/persistence/journal/JournalSpec.scala#L82), circe [ExportMacros.scala](https://github.com/circe/circe/blob/2fb611bb49619e4287b6ac048d2283c2781f4943/modules/generic-simple/src/main/scala/io/circe/generic/simple/util/macros/ExportMacros.scala#L4) and tapir [Main.scala](http://github.com/softwaremill/tapir/blob/6a0d3c349046c4357e9d53126ef4e7f0ac5dbee5/openapi-codegen/sbt-plugin/src/sbt-test/sbt-openapi-codegen/schema_inheritance/src/main/scala/Main.scala#L6).

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -95,6 +95,9 @@ module.exports = grammar({
   conflicts: $ => [
     [$.tuple_type, $.parameter_types],
     [$.inline_modifier, $._soft_identifier],
+    // 'extension' • '{' is either an extension definition with a braced body
+    // or the soft identifier `extension` (a Scala 2 id) called with a block.
+    [$.extension_definition, $._soft_identifier],
     [$.binding, $._simple_expression],
     [$.binding, $._type_identifier],
     [$.while_expression, $._simple_expression],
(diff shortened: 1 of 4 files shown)


eed3si9n (approved):
Thanks!

PR #574, opened 2026-07-24
fix: bare `$` identifier called with arguments

exoego:
- Corresponding issue: None.

`$(selector).submit()` misread the `$` as the start of a splice. The splice opening becomes the two-character tokens `${` and `$[`. A bare `$` before a parenthesis then lexes as an ordinary identifier. The dead `$ident` branch of splice_expression is removed because a `$ident` splice already lexes as one identifier.

Seen in playframework [Selenium.scala](https://github.com/playframework/playframework/blob/0fbcf712635e1f3fa84d6a6e97521efc39e13f84/testkit/play-test/src/main/scala/play/api/test/Selenium.scala#L41-L43)


eed3si9n (approved):
Thanks!

PR #575, opened 2026-07-24
fix: annotation arguments and multiple argument lists in extends clauses

exoego:

scalac reads annotation arguments greedily. In `extends B @Ann(x)` the `(x)` belongs to the annotation, not the constructor. The parser used to reduce the annotation empty and attach the arguments at the extends level. A following annotation such as `extends B @A(1) @B(2)` then broke.

Constructor argument lists now sit on _constructor_application as a repeat1 after a non-annotated simple type. extends_clause loses its own repeat(arguments). The repeat1 also fixes multiple argument lists followed by `with` or by comma-separated parents, as in `extends ValDef[T](a, b)(using NoSource) with W[T]`.

Annotation arguments seen in shapeless [annotation.scala](https://github.com/milessabin/shapeless/blob/99efca64f4f5/core/shared/src/test/scala/shapeless/annotation.scala#L50-L51). The multiple constructor argument lists are seen in scala3 [Trees.scala ](https://github.com/scala/scala3/blob/a036a3a/compiler/src/dotty/tools/dotc/ast/Trees.scala#L1089-L1090) and [untpd.scala](https://github.com/scala/scala3/blob/a036a3a/compiler/src/dotty/tools/dotc/ast/untpd.scala#L171).

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -887,8 +887,16 @@ module.exports = grammar({
           // but doing so increases the state of template_body to 4000
           $._structural_type,
           // This adds _simple_type, but not the above intentionally.
-          seq($._simple_type, field("arguments", $.arguments)),
-          seq($._annotated_type, field("arguments", $.arguments)),
+          // Constructor arguments attach only to a non-annotated simple type.
+          // After an annotation, argument lists belong to the annotation.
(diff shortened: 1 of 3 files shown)


eed3si9n (approved):
Thanks!

PR #576, opened 2026-07-25
perf: lex block comments in the external scanner

exoego:

Block comments were a multi-token nonterminal in extras. Their leading `/*` token occupied a column in nearly every parse-table row. They are now a single external BLOCK_COMMENT token with proper nesting (SLS 1.4). The comment-aware layout decisions (INDENT, OUTDENT, automatic semicolon) lex the comment at the points where the scanner previously deferred to the internal comment tokens. The decisions themselves are unchanged.

External extras are valid in every state. Positions where `/*` is ordinary text suppress the comment through the never-returned _suppress_block_comment token, offered as a dead grammar alternative inside line comments. String-middle states are excluded through their own valid_symbols.

The tree changes for nested comments. block_comment is now a leaf and nested comments no longer produce nested block_comment nodes.

```scala
  /* a /* b */ c */
```

- before: `(block_comment (block_comment))`
- after:  `(block_comment)`

Parsing the scala3 compiler corpus gets about 1.2x faster. The failure set on scala/scala 2.13 and on scala3 main is unchanged.

## 

This is a groundwork for an upcoming ASI fix that must look past a comment when deciding whether a newline ends a statement. For example,
```scala
  val x = 1
        /* c */ else 2
```

That requires consuming the comment as a token while carrying the preceding newline count over to the next scan.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -64,6 +64,13 @@ module.exports = grammar({
     "extends",
     "derives",
     "with",
+    // Lexed externally so the leading `/*` token does not occupy a column in
+    // nearly every parse-table row. That costs about 0.6MiB of parser.c.
+    $.block_comment,
+    // Never returned by the scanner. External extras look valid in every
+    // state, so the states where `/*` is plain text offer this token as a
+    // dead alternative and the scanner skips block comments there.
+    $._suppress_block_comment,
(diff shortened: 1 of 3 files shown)


eed3si9n (approved):
Thanks!

PR #580, opened 2026-07-26
fix: braceless match with cases at the same indentation

### LINKED ISSUE #264: Braceless `match` with cases of same indentation

exoego:
- Spin-out from #547
- Closes #264

Scala 3 lets the `case` clauses of a braceless `match` or catch align with the enclosing region instead of indenting one level deeper. The scanner now opens a flagged `case` region at that width. It closes the region on a later line at the same width that does not start another `case` clause.

Seen in scala3 [WConf.scala](https://github.com/scala/scala3/blob/a036a3a74024d97cc70b8d51f2612c05ec7ff881/compiler/src/dotty/tools/dotc/reporting/WConf.scala#L18-L24).

### DIFF
--- a/src/scanner.c
+++ b/src/scanner.c
@@ -148,6 +148,21 @@ static void advance_past_blanks(TSLexer *lexer) {
   }
 }
 
+// Marks an indent region whose `case` clauses align with their `match`/`catch`
+// (Scala 3 same-width case). Such a region also closes on a same-width line
+// that is not another `case` clause. The flag is packed into the int16 width.
+#define CASE_INDENT_FLAG ((int16_t)0x4000)
+
+static inline int16_t indent_width(int16_t entry) {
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Nice!

PR #581, opened 2026-07-26
fix: keep an internal comma from closing an indented block

exoego:
- spin-out from #547
- issue: None

A comma that continues its statement on the same line is internal to an indented block. Comma-separated parents, import lists, and enum cases all use one. The scanner now emits COMMA_OUTDENT only when the comma ends its line. A comma with code after it on the same line keeps the block open.

Seen in scala3 [Setup.scala](https://github.com/scala/scala3/blob/a036a3a74024d97cc70b8d51f2612c05ec7ff881/compiler/src/dotty/tools/dotc/cc/Setup.scala#L446).

### DIFF
--- a/src/scanner.c
+++ b/src/scanner.c
@@ -496,16 +496,46 @@ bool tree_sitter_scala_external_scanner_scan(void *payload, TSLexer *lexer,
     skip(lexer);
   }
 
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Thanks!

Issue #541, opened 2026-05-12
`if (...)` `then` with multiline condition parses incorrectly


iigaz:
### Commit of tree-sitter-scala you tested this on

a680000

### A code sample showing the error

```scala
if (false)
  || true
then
  print("true")
```

Also seen in:  https://github.com/scala/scala3/blob/65a6143a129929df825e3693209ead96de5a4e49/compiler/src/dotty/tools/dotc/typer/RefChecks.scala#L544, and https://github.com/scala/scala3/blob/65a6143a129929df825e3693209ead96de5a4e49/compiler/src/dotty/tools/dotc/typer/ProtoTypes.scala#L421, and https://github.com/scala/scala3/blob/65a6143a129929df825e3693209ead96de5a4e49/compiler/src/dotty/tools/dotc/cc/CaptureOps.scala#L196

### Show the error node

```
(compilation_unit [0, 0] - [4, 0]
  (if_expression [0, 0] - [2, 0]
    condition: (parenthesized_expression [0, 3] - [0, 10]
      (boolean_literal [0, 4] - [0, 9]))
    consequence: (indented_block [1, 2] - [2, 0]
      (postfix_expression [1, 2] - [1, 9]
        (operator_identifier [1, 2] - [1, 4])
        (identifier [1, 5] - [1, 9]))))
  (identifier [2, 0] - [2, 4])
  (call_expression [3, 2] - [3, 15]
    function: (identifier [3, 2] - [3, 7])
    arguments: (arguments [3, 7] - [3, 15]
      (string [3, 8] - [3, 14]))))
```

### What do you expect the tree to look like

```
(compilation_unit [0, 0] - [3, 0]
  (if_expression [0, 0] - [3, 0]
    condition: (infix_expression [0, 3] - [0, 18]
      left: (parenthesized_expression [0, 3] - [0, 10]
        (boolean_literal [0, 4] - [0, 9]))
      operator: (operator_identifier [0, 11] - [0, 13])
      right: (boolean_literal [0, 14] - [0, 18]))
    consequence: (indented_block [2, 2] - [3, 0]
      (call_expression [2, 2] - [2, 15]
        function: (identifier [2, 2] - [2, 7])
        arguments: (arguments [2, 7] - [2, 15]
          (string [2, 8] - [2, 14]))))))
```

### Where are you experiencing this error?

tree-sitter CLI

---

[...]

Issue #539, opened 2026-05-12
`else` part of if expression is parsed as an infix expression


susliko:
### Commit of tree-sitter-scala you tested this on

a680000

### A code sample showing the error

```scala
def a = if (true)
  1 else 2
```

### Show the error node

```
(compilation_unit [0, 0] - [2, 0]
  (function_definition [0, 0] - [2, 0]
    name: (identifier [0, 4] - [0, 5])
    body: (if_expression [0, 8] - [2, 0]
      condition: (parenthesized_expression [0, 11] - [0, 17]
        (boolean_literal [0, 12] - [0, 16]))
      consequence: (indented_block [1, 2] - [2, 0]
        (infix_expression [1, 2] - [1, 10]
          left: (integer_literal [1, 2] - [1, 3])
          operator: (identifier [1, 4] - [1, 8])
          right: (integer_literal [1, 9] - [1, 10]))))))
```

Issue #358, opened 2023-09-19
`for` with multiline expression enumerator


susliko:
### Commit of tree-sitter-scala you tested this on

59ab070df6667cad5bd669dd31aa866913129aa2

### A code sample showing the error

```scala
for 
  _ <- a *>
       b
yield ()
```

### Show the error node

```
(compilation_unit [0, 0] - [4, 0]
  (ERROR [0, 0] - [4, 0]
    (enumerators [1, 2] - [4, 0]
      (enumerator [1, 2] - [1, 11]
        (wildcard [1, 2] - [1, 3])
        (postfix_expression [1, 7] - [1, 11]
          (identifier [1, 7] - [1, 8])
          (operator_identifier [1, 9] - [1, 11])))
      (ERROR [2, 7] - [3, 8]
        (infix_pattern [2, 7] - [3, 8]
          left: (identifier [2, 7] - [2, 8])
          operator: (identifier [3, 0] - [3, 5])
          right: (tuple_pattern [3, 6] - [3, 8]
            (identifier [3, 7] - [3, 7])))))))
```

### What do you expect the tree to look like

```
(compilation_unit [0, 0] - [3, 0]
  (for_expression [0, 0] - [2, 8]
    enumerators: (enumerators [1, 2] - [2, 0]
      (enumerator [1, 2] - [1, 13]
        (wildcard [1, 2] - [1, 3])
        (infix_expression [1, 7] - [1, 13]
          left: (identifier [1, 7] - [1, 8])
          operator: (operator_identifier [1, 9] - [1, 11])
          right: (identifier [1, 12] - [1, 13]))))
    body: (unit [2, 6] - [2, 8])))
```

### Where are you experiencing this error?

vanilla tree-sitter

Issue #329, opened 2023-08-07
Error seen on parsing a case class nested in an object definition with Scala 3.


sageserpent-open:
### Commit of tree-sitter-scala you tested this on

dd02e510e92e057bdcadc4def1a3b28bed5fbea3

### A code sample showing the error

```scala
package com.sageserpent.kineticmerge.core

object Merge:

  case class Location(runIndex: Int, locationInRun: String):
  end Location

end Merge
```

### Show the error node

```
compilation_unit (0, 0) - (8, 0)
  package_clause (0, 0) - (0, 41)
    package (0, 0) - (0, 7) "package"
    package_identifier (0, 8) - (0, 41)
      identifier (0, 8) - (0, 11) "com"
      . (0, 11) - (0, 12) "."
      identifier (0, 12) - (0, 23) "sageserpent"
      . (0, 23) - (0, 24) "."
      identifier (0, 24) - (0, 36) "kineticmerge"
      . (0, 36) - (0, 37) "."
      identifier (0, 37) - (0, 41) "core"
  object_definition (2, 0) - (7, 0)
    object (2, 0) - (2, 6) "object"
    identifier (2, 7) - (2, 12) "Merge"
    template_body (2, 12) - (7, 0)
      : (2, 12) - (2, 13) ":"
      class_definition (4, 2) - (4, 59)
        case (4, 2) - (4, 6) "case"
        class (4, 7) - (4, 12) "class"
        identifier (4, 13) - (4, 21) "Location"
        class_parameters (4, 21) - (4, 59)
          ( (4, 21) - (4, 22) "("
          class_parameter (4, 22) - (4, 35)
            identifier (4, 22) - (4, 30) "runIndex"
            : (4, 30) - (4, 31) ":"
            type_identifier (4, 32) - (4, 35) "Int"
          , (4, 35) - (4, 36) ","
          class_parameter (4, 37) - (4, 58)
            identifier (4, 37) - (4, 50) "locationInRun"
            : (4, 50) - (4, 51) ":"
            type_identifier (4, 52) - (4, 58) "String"
          ) (4, 58) - (4, 59) ")"
      ERROR (4, 59) - (5, 14)
        : (4, 59) - (4, 60) ":"
  end (7, 0) - (7, 3) "end"
  _end_ident (7, 4) - (7, 9) "Merge"
```

### What do you expect the tree to look like

[...]

PR #582, opened 2026-07-26
chore: reserve _colon_eol and _operator_eol external tokens

exoego:
First of two PRs.
Closes no issue.

Declares two external tokens (`_colon_eol` for #329), `_operator_eol` for #541 / #358 / #539) without using them yet. A follow-up scanner PR emits and consumes them.

Declaring them first lets the sync workflow enlarge `parser.c` before the scanner reads the new slots. Otherwise a single scanner PR runs its new scanner against the stale checked-in `parser.c`, and the fuzz job hits an out-of-bounds read on `valid_symbols`. Appended at the end so existing token indices are unchanged.

Behavior-neutral (dead alternatives, like `error_sentinel`): `generate` raises `EXTERNAL_TOKEN_COUNT` 23→25 with order unchanged, tests pass 192/192, and parse trees are byte-identical to master across the smoke corpus.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -72,6 +72,12 @@ module.exports = grammar({
     // dead alternative and the scanner skips block comments there.
     $._suppress_block_comment,
     $.error_sentinel,
+    // Reserved for the follow-up scanner PR (a line-final `:` and a line-final
+    // symbolic operator). Declared here first so the generation bot enlarges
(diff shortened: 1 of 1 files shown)


eed3si9n (approved):
Thanks!

eed3si9n:
Generated in https://github.com/tree-sitter/tree-sitter-scala/pull/583

PR #585, opened 2026-07-27
chore: reserve the floating point separators external token

exoego:
Declare _floating_point_with_separators in the externals array, unreferenced for now. The follow-up scanner PR lexes a floating point literal whose integer part uses digit separators. Placed last so every existing external keeps its index. Behaviour is unchanged.

This will be used to parse `1_000.5` used in the Lila corpus.


eed3si9n (approved):
Thanks!

PR #586, opened 2026-07-27
feat: start a colon argument with a context function lambda

exoego:
A colon argument could start with a plain `x =>` lambda but not a context function `x ?=>` lambda. Use anyArrow so both forms open the argument. Without this the `?=>` fell into the block body and a dedented method chain after it failed to parse.

Seen in lila [Spotlight.scala,](https://github.com/lichess-org/lila/blob/cc7032e80ba7eae1b1b7640278513245d31c9e14/modules/tournament/src/main/Spotlight.scala#L28-L29) [SwissApi.scala](https://github.com/lichess-org/lila/blob/cc7032e80ba7eae1b1b7640278513245d31c9e14/modules/swiss/src/main/SwissApi.scala#L172-L173), [PuzzleBatch.scala ](https://github.com/lichess-org/lila/blob/cc7032e80ba7eae1b1b7640278513245d31c9e14/modules/puzzle/src/main/PuzzleBatch.scala#L28-L30)and [TournamentApi.scala](https://github.com/lichess-org/lila/blob/cc7032e80ba7eae1b1b7640278513245d31c9e14/modules/tournament/src/main/TournamentApi.scala#L293-L295).

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -1710,7 +1710,9 @@ module.exports = grammar({
                   $._identifier,
                   $.wildcard,
                 ),
-                fatArrow(),
+                // anyArrow so a context-function lambda `x ?=>` also starts a
+                // colon argument, not only the plain `=>` form.
+                anyArrow(),
               ),
             ),
           ),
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Thanks!

PR #588, opened 2026-07-27
feat: lex floating point literals with digit separators

exoego:
> [!WARNING]
> This requires a parser genration to include #585

A digit separator in the integer part of a float such as `99_999.9999` could not be lexed. The internal number regex drops the dot transition after an underscore, so the scanner lexes that shape as _floating_point_with_separators. The fraction and exponent gain separators through the regex.

Seen in lila [PlanPricingTest.scala](https://github.com/lichess-org/lila/blob/cc7032e80ba7eae1b1b7640278513245d31c9e14/modules/plan/src/test/PlanPricingTest.scala#L35-L36).

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -77,16 +77,8 @@ module.exports = grammar({
     // dead alternative and the scanner skips block comments there.
     $._suppress_block_comment,
     $.error_sentinel,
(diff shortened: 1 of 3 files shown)


eed3si9n (approved):
Thanks!

PR #589, opened 2026-07-27
feat: bind an indented block to a for comprehension enumerator

exoego:
A generator or `=` binding whose right side is a braceless multi-statement block failed to parse. Allow an indented block there, as the guard and the for body already do.

Seen in lila [MovePlayer.scala,](https://github.com/lichess-org/lila/blob/cc7032e80ba7eae1b1b7640278513245d31c9e14/modules/round/src/main/MovePlayer.scala#L94-L97), [AutoPairing.scala](https://github.com/lichess-org/lila/blob/cc7032e80ba7eae1b1b7640278513245d31c9e14/modules/tournament/src/main/AutoPairing.scala#L41) and etc.

Lila is now 100% covered 😎 

> ok, lila/modules/: 100.00%, expected at least 99%

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -2326,7 +2326,9 @@ module.exports = grammar({
           optional("case"),
           $._pattern,
           choice("<-", "←", "="),
-          $.expression,
+          // A generator or `=` binding can bind a braceless indented block, as
+          // the guard and the `for` body already do.
+          choice($.expression, $.indented_block),
           optional($.guard),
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Thanks!

eed3si9n:
Updated the smoke test values here - https://github.com/tree-sitter/tree-sitter-scala/pull/590

PR #591, opened 2026-07-28
perf: inline small hidden rules to shrink parser.c

exoego:
- spin-out from #547 

Add eight small hidden rules to the inline list. Each reduces to a token almost at once, so inlining removes the reduce step and merges states. This drops parser.c by about 2MB and LARGE_STATE_COUNT from 4182 to 2910. The tree is unchanged (node-types.json identical). early_defs becomes a visible rule because it can no longer alias the now-inlined _braced_template_body.

- parser.c: 30.34 MiB -> 28.28 MiB (-2.06 MiB, -6.8%)
- LARGE_STATE_COUNT: 4182 -> 2910 (-1272, -30%)
- No speed change

This creates a room for new big features (capture checking ? maybe).

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -93,6 +93,16 @@ module.exports = grammar({
     $._param_value_type,
     $._simple_type,
     $.literal,
+    // Small hidden rules that reduce to a token almost immediately. Inlining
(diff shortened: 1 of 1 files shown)


eed3si9n (approved):
Thanks!

PR #593, opened 2026-07-28
fix: parse a lambda parameter with a bare function type

### LINKED ISSUE #543: lambda that accepts a function parses as an infix expression

a680000

```scala
val test = (f: Int => Int) => f(1)
```

Also seen in: https://github.com/scala/scala/blob/0929a0120f78eb384396b63eb954b65a2288fd07/src/compiler/scala/tools/nsc/transform/patmat/MatchCodeGen.scala#L114

### Show the error node

```
  (val_definition [0, 0] - [0, 34]
    pattern: (identifier [0, 4] - [0, 8])
    value: (infix_expression [0, 11] - [0, 34]
      left: (parenthesized_expression [0, 11] - [0, 26]
        (lambda_expression [0, 12] - [0, 25]
          parameters: (identifier [0, 12] - [0, 13])
          parameters: (type_identifier [0, 15] - [0, 18])
          (identifier [0, 22] - [0, 25])))
      operator: (operator_identifier [0, 27] - [0, 29])
      right: (call_expression [0, 30] - [0, 34]
        function: (identifier [0, 30] - [0, 31])
        arguments: (arguments [0, 31] - [0, 34]
          (integer_literal [0, 32] - [0, 33]))))))
```

Wrapping `Int => Int` into parentheses gives the correct tree:

```
  (val_definition [0, 0] - [0, 36]
    pattern: (identifier [0, 4] - [0, 8])
    value: (lambda_expression [0, 11] - [0, 36]
      parameters: (bindings [0, 11] - [0, 28]
        (binding [0, 12] - [0, 27]
          name: (identifier [0, 12] - [0, 13])
          type: (tuple_type [0, 15] - [0, 27]
            (function_type [0, 16] - [0, 26]
              parameter_types: (parameter_types [0, 16] - [0, 19]
                (type_identifier [0, 16] - [0, 19]))
              return_type: (type_identifier [0, 23] - [0, 26])))))
      (call_expression [0, 32] - [0, 36]
        function: (identifier [0, 32] - [0, 33])
        arguments: (arguments [0, 33] - [0, 36]
          (integer_literal [0, 34] - [0, 35]))))))
```

tree-sitter CLI

exoego:
- Closes #543

A lambda parameter typed with a bare function type such as `(f: Int => Int) => f(1)` was misread as a parenthesized expression joined by an infix `=>`. Let the lambda parameter clause win so it parses as a lambda whose parameter has a function type.

Seen in scala3 [Inliner.scala](https://github.com/scala/scala3/blob/a036a3a74024d97cc70b8d51f2612c05ec7ff881/compiler/src/dotty/tools/dotc/inlines/Inliner.scala#L433), scala [MatchCodeGen.scala](https://github.com/scala/scala/blob/e2bf3551286f729b4f2bec78b36e3f890fccfd80/src/compiler/scala/tools/nsc/transform/patmat/MatchCodeGen.scala#L114), zio [ZIO.scala ](https://github.com/zio/zio/blob/3dba37da7d38ae967ffe4817086af2520a63eb71/core/shared/src/main/scala/zio/ZIO.scala#L3866)and tapir [VertxZioServerInterpreter.scala](https://github.com/softwaremill/tapir/blob/11b5ed5da5fc0196a645b77b12941194cf355ebd/server/vertx-server/zio/src/main/scala/sttp/tapir/server/vertx/zio/VertxZioServerInterpreter.scala#L60).

---
Why dynamic: https://github.com/tree-sitter/tree-sitter-scala/pull/593#discussion_r3669959038

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -125,6 +125,12 @@ module.exports = grammar({
     [$.extension_definition, $._soft_identifier],
     [$.binding, $._simple_expression],
     [$.binding, $._type_identifier],
+    // A parenthesized type before `=>` can extend into a function type or stop
+    // so the arrow belongs to the enclosing lambda, pattern, or colon argument.
+    // The conflicts keep both readings alive so precedence picks per context.
+    [$._type, $.parameter_types],
+    [$.parameter_types, $._colon_bindings],
+    [$.parameter_types, $.bindings],
     [$.while_expression, $._simple_expression],
     [$.if_expression],
     [$.match_expression],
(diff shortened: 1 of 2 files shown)


grammar.js
```diff
@@ -1311,28 +1317,30 @@ module.exports = grammar({
...
 
+    // A plain type wins where both readings are legal (ascription, pattern,
+    // colon argument). In a lambda parameter clause a plain type leaves the
+    // arrow dangling, so the function type is the only survivor there.
     function_type: $ =>
-      prec.left(
-        choice(
-          seq(field("type_parameters", $.type_parameters), $._arrow_then_type),
-          seq(field("parameter_types", $.parameter_types), $._arrow_then_type),
+      prec.dynamic(
```
eed3si9n, line 1324: Do you know why three `prec.dynamic(...)`s are required in this PR? Is the syntax ambiguous with something else?

exoego: Take `(f: Int => Int) => f(1)`. The correct Scala parse is a lambda whose parameter f has the function type `Int => Int`. However, tree-sitter-scala also admits a second, wrong parse. It treats `(f: Int => Int)` as a parenthesized expression and the `=>` after ) as an infix operator. That is not valid Scala, since `=>` is reserved and never an operator. The grammar allows it only because its `operator_identifier` token is built from operator characters and does not exclude `=>`. Before the fix, tree-sitter picked this misparse, so the bug is a grammar misinterpretation, not a real language ambiguity.

A static rule requires much effort to fix it. The deciding point is whether `Int => Int` is a function type, and that part of the type grammar is shared across bindings, ascriptions, patterns, and colon arguments. So, one static setting hits them all. I confirmed both directions fail. Preferring the plain type brings the bug back, and preferring the function type breaks `case x: Int => body`, ascriptions, and braced lambdas. So the fix keeps both parses alive with a conflict and ranks them with dynamic precedence, which is compared on the finished trees. That lets the valid reading win per context.

The cost is about +3.5 KB of parser size and roughly a 3 percent parse slowdown (on my local huge pekko/akka corpus). So I think this is accepetable.

eed3si9n (approved):
Thanks!

PR #596, opened 2026-07-30
feat(scala3): attach end markers to the construct they close

### LINKED ISSUE #275: `end` clause capture/grouping

236c7b2

```scala
object C:
  given C =
    new C:
      def f = "!"
      end f
    end new
  end given
end C
```

### Show the error node

```
(object_definition object body: (identifier)
  (template_body :
   (given_definition given return_type: (type_identifier) =
    body: 
     (instance_expression new arguments: (type_identifier)
      (template_body :
       (function_definition def name: (identifier) = body: (string))
       end _end_ident)))
   end new end given))
 end _end_ident
```

There is no ERROR node, but I strongly believe that the `end` clause of a given node should be captured as a child of that node (or some similar construct).
Particularly, looking at the second last line of the output, we can see that `end new` and `end given` are siblings. Meanwhile, the code suggests that `new` block is a child of the `given` block. 
The improper grouping of the `end` clauses makes it so that a lot more of the parse tree has to be examined to find the proper indentation for that `end` clause.

For instance, you have to check if you're not already correctly indented by checking if the previous sibling is a matching block to the `end` clause. But since multiple nested `end` clauses are just represented as sibling nodes, you have to walk through all the previous children, and then you go one level up and so on.

treesit (GNU Emacs 29+)

### LINKED ISSUE #511: Named end-markers

a06047ee441bd02ca6ebcc0f913737b7c6e178e1

```
class Foo:
  println("")
end Foo
```

### Show the error node

There is no error node. But the grammar allows standalone end markers, not actually ending any definition, which is incorrect.

Image

_No response_

_No response_

exoego:
- Fixes #275
- Fixes #511

An `end` marker was a hidden standalone statement. Nested markers became siblings of the definition they close. A lone `end` marker parsed silently anywhere.
Now the marker is the named `end_marker` node. It is the last child of the definition or expression it closes. Each construct admits only the tags that can close it. So `end match` goes to the match expression and `end foo` goes to the enclosing definition. A standalone `end` marker stays an ordinary expression.

A naive slot would let the `end` keyword directly follow any expression. The parse table then duplicates the expression and type states for those contexts.
To avoid this, a slot starts with the automatic semicolon. The scanner emits the `end` keyword only right after that semicolon, and only when the line looks like a marker. So the `end` keyword never directly follows an expression. The scanner also stops offering the semicolon where no statement can end. Parenthesized `if` and loop heads and braced match bodies take no marker. Real code (34 Scala OSSs on my end) only marks the indentation forms, and the extra forks could exceed the GLR version limit. The new states re-roll some GLR ties. Explicit dynamic precedences pin them.

Three side effects.

1. A `case` clause now ends at its content instead of at the next clause.
2. The `while` condition tolerates the new semicolon before its `do`.
3. The state changes re-rolled GLR ties that had hidden real misreadings in the lambda rules. Fixing those needed a split of the lambda parameter rules. A braced lambda keeps its unparenthesized typed parameter, which dotty itself only accepts in block position, and an expression keeps the colon argument reading. parser.c ends up smaller than on master (-300KB).

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -41,6 +41,16 @@ const ascriptionArrowTail = $ =>
 // body, and a do-while fork there doubles the paren-for automaton (~+7MiB).
 const statementExpression = $ => choice($.expression, $.do_while_expression);
 
+// A Scala 3 end marker as the last child of the construct it closes. The
+// leading semicolon keeps `end` out of every expression follow set, and
+// the weight beats the statement reading of the marker line. The tails are
+// shared rules so every construct reuses the same post-semicolon states.
+const endMarkerTail = ($, marker, weight) =>
+  prec.dynamic(
+    weight,
+    seq($._automatic_semicolon, alias(marker, $.end_marker)),
+  );
+
 module.exports = grammar({
   name: "scala",
 
@@ -112,8 +122,6 @@ module.exports = grammar({
   // These names can be used in the prec functions to define precedence relative only to other names in the array, rather than globally.
   precedences: $ => [
     ["mod", "soft_id"],
-    ["end", "soft_id"],
-    ["end", "this_id"],
     ["new", "structural_type"],
     ["self_type", "lambda"],
     ["annotation", "applied_constructor_type"],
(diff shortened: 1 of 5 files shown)


eed3si9n:
Thanks for the contribution!

grammar.js
```diff
@@ -485,13 +509,16 @@ module.exports = grammar({
...
+      seq(
+        prec.left(
+          seq(
+            field("name", $._identifier),
+            field("extend", optional($.extends_clause)),
+            field("derive", optional($.derives_clause)),
+            field("body", optional($._definition_body)),
+          ),
         ),
+        optional($._end_marker_named_tail),
```
eed3si9n, line 521: Note to self: The nested `seq(...)` here allows `$._end_marker_named_tail` to use the dynamic precedence instead of the static precedence.

eed3si9n (approved):
Thanks!

Issue #218, opened 2023-05-23
Incorrect precedences of binary logical operators


susliko:
### Commit of tree-sitter-scala you tested this on

2d0e6b8

### A code sample showing the error

```scala
object X {
  a < b && c > d
}
```

### Show the error node

```
object_definition [0, 0] - [2, 1]
  name: identifier [0, 7] - [0, 8]
  body: template_body [0, 9] - [2, 1]
    infix_expression [1, 2] - [1, 16]
      left: infix_expression [1, 2] - [1, 12]
        left: infix_expression [1, 2] - [1, 7]
          left: identifier [1, 2] - [1, 3]
          operator: operator_identifier [1, 4] - [1, 5]
          right: identifier [1, 6] - [1, 7]
        operator: operator_identifier [1, 8] - [1, 10]
        right: identifier [1, 11] - [1, 12]
      operator: operator_identifier [1, 13] - [1, 14]
      right: identifier [1, 15] - [1, 16]
```

### What do you expect the tree to look like

Logical binary operators (&&, ||) should have higher precedence over (<, >, ==, <=, >=)
```
object_definition [0, 0] - [2, 1]
  name: identifier [0, 7] - [0, 8]
  body: template_body [0, 9] - [2, 1]
    infix_expression [1, 2] - [1, 16]
      left: infix_expression
        left: identifier
        operator: operator_identifier (<)
        right: identifier
      operator: operator_identifier (&&)
      right: infix_expression
        left; identifier
        operator: operator_identifier (>)
        right: identifier
```

### Where are you experiencing this error?

nvim-treesitter

PR #600, opened 2026-07-31
perf: shrink parser.c from 29.4MB to 18.3MB with a control-tail gate

exoego:
- This is a preparation for #218 which enlarges parser.c a lot.

A new zero-width external token sits right before catch, finally, else, then, and yield where the grammar allows one. Every construct body then ends with the same single gate column instead of its own keyword follow set, so the expression automaton stops being cloned per enclosing construct. The scanner emits the gate where the next word is one of the gated keywords. do and while stay ungated because they can start a statement and gate validity blends in from unrelated GLR forks.

Verified over 14,980 corpus files across 35 repos on my end. Trees are byte identical except one file where the parser previously split a curried fold block application into two statements and now reads it as scalac does, and one already failing file whose error recovery shape-shifts without changing error counts.


eed3si9n (approved):
Thanks!

PR #601, opened 2026-07-31
fix: rest-of-block lambdas and comma-closed templates in indented blocks

exoego:
Two gaps where indented bodies lagged behind braces.
1. A lambda statement in an indented block now takes the remaining statements as its body. Seen in Akka
2. A dedented comma can close an indented `new:` template used as anargument. Seen in Lila


eed3si9n (approved):
Thanks!

PR #606, opened 2026-08-01
feat: support XML literals and XML patterns 

exoego:
The `<` that opens a literal is lexed by the new external token XML_TAG_START. The scanner returns it only where the grammar admits a literal and the `<` is immediately followed by a name start character, so `a < b`, a leading infix `< b`, `<-` and `<:` keep lexing as operators. Comment, CDATA and processing instruction literals need no scanner support because their tokens beat the operator tokens by length.

Expression elements and pattern elements share the open tag rule, so a position where both are viable shifts the same states until the element body decides. Top level patterns inside an embed are a restricted list. An alternative whose left edge is the full pattern rule drags given patterns and ascriptions into the same closure as the statement grammar of a block, and the conflicts then cascade beyond resolution. `_*` is kept through a restricted repeat pattern alias, and nested pattern positions still take any pattern.

- Seen in [scala-xml](https://github.com/scala/scala-xml/blob/a4c4e52e88ea92fb019650208984170942407db4/shared/src/test/scala/scala/xml/PatternMatchingTest.scala#L28)
- Seen in [akka](https://github.com/akka/akka/blob/e75e809380ee16a18b12007d69bb46c30b9230e7/akka-stream-tests/src/test/scala/akka/stream/scaladsl/CoupledTerminationFlowSpec.scala#L35)
- Seen in [play](https://github.com/playframework/playframework/blob/aed7df209fc4074147408b7ea16b5259ec3faed8/documentation/manual/working/scalaGuide/main/xml/code/ScalaXmlRequests.scala#L45)
- Seen in [kamon](https://github.com/kamon-io/Kamon/blob/a376f39873921a1b2def82942e06950756522307/project/Build.scala#L239)

----
This is the last PR in the series for the new features.
Publishing new version is appreciated.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -165,6 +165,9 @@ const fatArrow = () => choice("=>", alias("⇒", "=>"));
 // `=>` or the context-function arrow `?=>`.
 const anyArrow = () => choice(fatArrow(), "?=>");
 
+// XML Name (SLS §10 / XML spec), covering namespaced names like `x:ga`.
+const XML_NAME = /[_\p{L}][-.:_\p{L}\p{Nd}]*/;
+
 // Accepts a `:` in both spellings. The scanner lexes a line-final lone colon
 // as the external COLON_EOL token (scalac's COLONeol), so every consumer of
 // such a colon must handle it.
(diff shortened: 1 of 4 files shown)


eed3si9n (approved):
Thanks!

PR #607, opened 2026-08-01
perf: cut parse time by a third 

exoego:
## What

Five changes to remove hot parse path. No grammar rule changes shape, and no node type changes.

## Numbers

Measured on a 35 repository corpus (14979 files). Each figure is the best of four interleaved runs of a native harness that links the tree-sitter runtime and reuses one parser, so process startup and file IO stay out of the measurement.

| corpus | before | after | change |
| --- | ---: | ---: | ---: |
| scala (Scala 2) | 867 ms | 588 ms | −32.2 % |
| scala3 | 1000 ms | 687 ms | −31.4 % |
| lila | 872 ms | 560 ms | −35.8 % |
| all 35 repos | 10016 ms | 6594 ms | −34.2 % |

### Scanner

* The newline recovery test called `get_column`, which rescans the line from its start, on almost every token. The recovery adds nothing when `last_newline_count` is zero, so the whole test is now gated on that. This is the single largest win in the PR.
* `iswspace`, `iswalpha` and `iswalnum` are real library calls and the scanner runs them on every token. No locale is ever installed here, so the C locale sets they resolve to are spelled out inline. Deserialize also copies the indent stack in one `memcpy`.

### Grammar

* Plain names shift in place of reducing. The `identifier` rule wrapped a single token, so every name paid a unit reduction to build the node. Spelling the alternatives out with the alias on each one moves the node name onto the parent production.
* `_exprs_in_parens` is inlined. The vararg tail now shares the argument list repeat instead of branching into its own rule, which removes the conflict that blocked inlining. The Scala 2 vararg carries a dynamic weight because the ascription reading of the same text used to win only by the tie break that the restructure changed.
* `_definition_pattern` is inlined. The added conflicts are the ones the generator asks for once the rule stops absorbing them.
* Three conflict declarations the inlined rules no longer need are removed.

Note) Inlining `_infix_operand` was worth another 4 percent and is left out. It raises the `infix_expression` state count from 947 to 1939, over the 1800 ceiling that `script/smoke_test.sh` enforces.

## Parser size

`src/parser.c` grows from 23 MiB to 26.51 MiB and the state count from 14704 to 16783.. That is the price of the four inlined rules. Two further inline candidates were measured at about 1 percent for another 2 MiB and left out.

I think 35% boost justifies this growth.


eed3si9n (approved):
Thanks!

PR #608, opened 2026-08-01
feat: parse binary integer literals

exoego:
> [!NOTE]

Dotty's scanner reads `0b` and `0B` as a base 2 prefix. The grammar only had the hex prefix, so `0B0000` lexed as `0` followed by an identifier. The scanner loops over separators and rejects only a trailing one, so `0b0001__0000` is legal and the rule allows repeated underscores.

Seen in https://github.com/scala/scala3/blob/a036a3a74024d97cc70b8d51f2612c05ec7ff881/tests/run/binaryLiterals.scala


eed3si9n (approved):
Thanks!

PR #609, opened 2026-08-01
feat: parse a type with more than one refinement

exoego:
> [!NOTE]

Dotty's refinedTypeRest calls itself on the tree it just built, so a type takes any number of refinements. The grammar allowed only one, so `C { type U = T } { type T = String }` did not parse.

The rule recurses on compound_type instead of repeating the refinement. That nests the tree the way dotty nests RefinedTypeTree, and it costs 33KB less in parser.c than a repeat does.

A `#` projection still needs the parens that dotty needs, because dotty reads `#` in simpleTypeRest, which sits below refinement.

Seen in https://github.com/scala/scala3/blob/a036a3a74024d97cc70b8d51f2612c05ec7ff881/tests/pos/lookuprefined.scala Seen in https://github.com/scala/scala3/blob/a036a3a74024d97cc70b8d51f2612c05ec7ff881/tests/pos/projections.scala

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -1465,7 +1465,14 @@ module.exports = grammar({
             repeat1(seq("with", field("extra", $._annotated_type))),
           ),
         ),
-        prec.left(seq(field("base", $._annotated_type), $._refinement)),
+        // Dotty's refinedTypeRest recurses on itself, so a type takes any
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Thanks!

PR #610, opened 2026-08-01
feat: parse type bindings in a quote type block

exoego:
Dotty parses the body of `'[...]` as TypeBlock, which is a run of type definitions followed by the type itself. The rule took a bare type, so `'[type x <: AnyKind; x]` did not parse.

The separator is _semicolon rather than a literal `;` because dotty takes a newline there too, and a test in the corpus writes each binding on its own line. Requiring a literal `;` costs 105KB more and still misses that file. Leaving the separator optional costs 682KB more, because the context bound repetition stops closing on sight of the next `:`.

Seen in https://github.com/scala/scala3/blob/a036a3a74024d97cc70b8d51f2612c05ec7ff881/tests/pos-macros/multiline-quote-patterns.scala
Seen in https://github.com/scala/scala3/blob/a036a3a74024d97cc70b8d51f2612c05ec7ff881/tests/pos-macros/i7264.scala

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -2409,7 +2409,9 @@ module.exports = grammar({
           choice(
             // `'{}` is a quoted empty block, common in macro code.
             seq("{", optional($._block), "}"),
-            seq("[", $._type, "]"),
+            // TypeBlock ::= {TypeBlockStat semi} Type. The bindings let a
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Thanks!

PR #611, opened 2026-08-01
ci: measure the dotty test suite and drop non-Scala inputs

exoego:
I've noticed that dotty coverage does not improve for the last three PRs.
It turned out that smoke test does not cover dotty test source which contains lots of uncovered valid features.

This PR adds dotty's test source, except neg/invalid/disabled/pending/fuzzy ones.
It also excludes Scala REPL logs, which are not legit Scala source.


eed3si9n (approved):
🎉

PR #613, opened 2026-08-02
feat: allow a semicolon after a match type case

exoego:
> [!NOTE]

Dotty's typeCaseClause takes one optional `;` after the case body, which its grammar comment spells as `[semi]`. The rule had no separator, so `X match { case Int => Long ; case Long => Int }` did not parse.

The separator sits inside type_case_clause the way dotty consumes it, so a trailing `;` before the closing brace works too. Dotty reads at most one, and `;;` stays a syntax error here as well.

Seen in https://github.com/scala/scala3/blob/a036a3a74024d97cc70b8d51f2612c05ec7ff881/tests/pos/case-semi.scala Seen in https://github.com/scala/scala3/blob/a036a3a74024d97cc70b8d51f2612c05ec7ff881/tests/pos/i13331.scala

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -1590,7 +1590,14 @@ module.exports = grammar({
     type_case_clause: $ =>
       prec.left(
         PREC.control,
-        seq("case", $._infix_type_choice, field("body", $._arrow_then_type)),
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Thanks!

PR #614, opened 2026-08-02
feat: parse the erased modifier

exoego:
> [!NOTE]

`erased` marks a definition or a parameter whose value is dropped after type checking. It reaches four places. modifiers covers `erased def` and `erased val`, parameter covers `def f(erased x: Int)`, name_and_type covers the function type `(erased ct: CanThrow[E]) ?=> R`, and binding covers the lambda `(erased ev, x) => x + 2`.

The rule carries no "mod" precedence, unlike its neighbours, because `erased` is also a plain name. A static win would take the modifier in `def f(erased: Int)` and never recover. The declared conflict against _soft_identifier lets the token after it decide, which is what dotty does with its lookahead for a colon.

Seen in https://github.com/scala/scala3/blob/a036a3a74024d97cc70b8d51f2612c05ec7ff881/tests/pos/erased-soft-keyword.scala
Seen in https://github.com/scala/scala3/blob/a036a3a74024d97cc70b8d51f2612c05ec7ff881/tests/pos/erased-lambda.scala

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -350,6 +350,9 @@ module.exports = grammar({
     [$._infix_operand, $.vararg],
     [$.tuple_type, $.parameter_types],
     [$.inline_modifier, $._soft_identifier],
+    // 'erased' at a parameter start is the modifier when a name follows and the
+    // name itself when a colon does, which only the next token settles.
+    [$.erased_modifier, $._soft_identifier],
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Thanks!

PR #615, opened 2026-08-02
chore: reserve the erased modifier external token

exoego:
To optimize parser.c size for `erased` feature


eed3si9n (approved):
Thanks!

eed3si9n:
Synced in https://github.com/tree-sitter/tree-sitter-scala/pull/617

PR #616, opened 2026-08-02
feat: parse a given import selector outside braces

exoego:
A bound given selector is a WildCardSelector in dotty, so it stands alone as well as inside braces. The rule for it was only wired into the braced form, so `import bar.given Foo` did not parse while `import a.{given Foo}` did. The grammar comment already carried the right definition.

Seen in https://github.com/scala/scala3/blob/a036a3a74024d97cc70b8d51f2612c05ec7ff881/tests/run/i18183.given.scala
Seen in https://github.com/scala/scala3/blob/a036a3a74024d97cc70b8d51f2612c05ec7ff881/tests/warn/i20860.scala

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -658,6 +658,9 @@ module.exports = grammar({
               seq(
                 ".",
                 choice(
+                  // A bound given selector is a WildCardSelector, so it stands
+                  // on its own as well as inside braces.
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Thanks!

PR #618, opened 2026-08-03
perf: lex the erased modifier in the scanner, cut parser.c by 1MB

exoego:
`erased` is an ordinary name (`def erased(x: Int)`, `nme.erased`), so the keyword token had to be handed back through `_soft_identifier` in every name position, plus a declared conflict to pick between the two readings. It costed 1MB of parser.c

Now the scanner emits the modifier when a name or a definition keyword follows. `else`, `catch` and `finally` are read in the same arm. Both arms consume the word, so reading it once with read_word keeps them from eating each other's input. Splitting them let `else ""` lose its OUTDENT and `erased val` inside braces lose its modifier.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -206,6 +206,9 @@ const nameChoice = $ =>
     alias("super", $.identifier),
   );
 
+// The scanner-lexed `erased`, keeping the keyword token's node name.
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Thanks!

PR #619, opened 2026-08-03
feat: take a refinement as an intersection operand

exoego:
SimpleType1 admits a bare Refinement, so `{ type A = Int }` is a type of its own and belongs on either side of an infix type operator. The operands left that case out, so `A & { type X = Int }` did not parse. It covers 40 files in the corpus, most of them capture-checking sources.

The operand takes the braced spelling only, the way simpleType1 calls refinement(indentOK = false). Admitting the indented one costs 1.3MB of parser.c instead of 438KB.

Seen in https://github.com/scala/scala3/blob/a036a3a74024d97cc70b8d51f2612c05ec7ff881/tests/pos/i2858.scala
Seen in https://github.com/scala/scala3/blob/a036a3a74024d97cc70b8d51f2612c05ec7ff881/tests/pos/i5699.scala

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -352,6 +352,8 @@ module.exports = grammar({
     // ascribed argument or shift the colon of a Scala 2 vararg `x: _*`.
     [$._infix_operand, $.vararg],
     [$.tuple_type, $.parameter_types],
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Thanks!

PR #620, opened 2026-08-03
chore: drop two unnecessary conflicts

exoego:
generate warned that `_constructor_application` and `instance_expression` no longer conflict. `[$._simple_expression, $._type_identifier]` was also listed twice.

Removing both leaves parser.c byte-for-byte the same and the corpus failure set unchanged, so nothing depended on the GLR forks they asked for. Every other declaration was checked the same way.


eed3si9n (approved):
Thanks!

PR #621, opened 2026-08-03
feat: parse named type arguments

exoego:
Scala 3 takes a named type argument the way it takes a named term argument, so `f[T = Any](x)` and `Foo[K = Int, V = String]` are both legal.

Seen in https://github.com/scala/scala3/blob/a036a3a74024d97cc70b8d51f2612c05ec7ff881/tests/pos/named-typeargs.scala
Seen in https://github.com/scala/scala3/blob/a036a3a74024d97cc70b8d51f2612c05ec7ff881/tests/pos/i3955.scala


eed3si9n (approved):
Thanks!

PR #623, opened 2026-08-03
feat: parse the Scala 3.6 given syntax

exoego:
A given takes a template body straight after the type, so `given Monad[Foo] { ... }` needs no `with`, and a given type is an infix chain, which is what makes `given Int is FromString` one rather than a special form. A literal type could not be a given type either, and a structural instance leaves the body optional, so `given A() with B` names two constructors and stops. A conditional clause carries `tracked val`. None of it parsed. Together they cover 105 files in the corpus.

The type and body are one rule. Spelled inline, every state the type and body choices need is built twice, once behind the optional signature and once without it, which costs 792KB of parser.c on its own.

The infix and literal readings sit below the plain type on a dynamic weight. Without that, `given (using a: Int): Int = a` reads as a named tuple joined by an infix `:`, which parses with no error and builds the wrong tree.

A refinement is a constructor everywhere else, so `with {}` would read the brace as a further constructor rather than an empty body. The extras after the first constructor therefore leave refinements out, which is the only position where a brace is unambiguous.

Only the tracked modifier is admitted to a parameter, not the whole modifiers rule. Admitting all of them reads 14 further files wrong, `implicit` and `inline` in a parameter clause having other meanings already.

Seen in https://github.com/scala/scala3/blob/a036a3a74024d97cc70b8d51f2612c05ec7ff881/tests/pos/FromString.scala
Seen in https://github.com/scala/scala3/blob/a036a3a74024d97cc70b8d51f2612c05ec7ff881/tests/pos/multi-given.scala Seen in https://github.com/scala/scala3/blob/a036a3a74024d97cc70b8d51f2612c05ec7ff881/tests/pos/i13580.scala


eed3si9n (approved):
Thanks!

PR #625, opened 2026-08-04
chore: reserve the soft modifier external tokens

exoego:
Reserves ten tokens for new features and parser optimization.


eed3si9n (approved):
Thanks!

PR #627, opened 2026-08-05
perf: read the soft modifiers in the scanner, shrink parser.c by 2.76 MiB

exoego:
`open`, `opaque`, `infix`, `tracked`, `transparent`, `inline` and `into` are names as well as modifiers, so each one needed a soft-identifier alternative in every name position. The tables were paying about 150KB per word, and far more for `inline`, which also stands at the start of an expression and made the parser fork on every occurrence.

They join the scanner arm that already reads `erased`, which is now a table rather than a chain of comparisons. `inline` needs a follow test of its own, the way the reference parser uses canStartExprTokens. A name, a literal or an open bracket makes it the modifier. Across a line break only a reserved word counts, since a line starting with a soft modifier would otherwise capture a preceding bare `inline` that is really a value.

The soft identifier rule is left holding `extension` alone. The stale `end` alternative goes with the rest, since `end` has been scanner-lexed for a while and nothing matched it as a literal any more. Both queries asked for an `end` token that therefore never existed, so end markers were never highlighted. They ask for the marker node now.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -206,8 +206,16 @@ const nameChoice = $ =>
     alias("super", $.identifier),
   );
 
-// The scanner-lexed `erased`, keeping the keyword token's node name.
+// The scanner-lexed soft modifiers, each keeping its keyword token's node name.
 const erasedMod = $ => alias($._erased_modifier, $.erased_modifier);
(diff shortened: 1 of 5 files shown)


eed3si9n (approved):
Thanks!

PR #629, opened 2026-08-05
feat: parse the capture checking syntax

exoego:
A capture set rides on a type. `IterableOnce[A]^` is the short form of `^{cap}`, and `View[A]^{this}` names what the type captures. The set also stands where a type argument is expected, as in `foo[{io}](xs)`, and a type parameter carries one in `Stepper[_]^`.

`->` and `?->` are pure function arrows. They take a capture set of their own in `A ->{x} B`, and a by-name parameter reads as `msg: -> String`. They stay two literals rather than one token with a lexical precedence, because that precedence outranks match length and a combined arrow would take the head of an operator like `->>`.

`update` and `consume` are modifiers, and `uses` opens a second parent list. All three are ordinary names as well, so the scanner lexes them the way it already lexes the other soft modifiers. A modifier never precedes a word that continues an expression, which is what keeps `update match` a match on a parameter called `update`.


eed3si9n (approved):
Nice!

eed3si9n:
@amaanq Could you add @exoego to the tree-sitter-scala committer please?
https://github.com/tree-sitter/tree-sitter-scala/graphs/contributors?from=8%2F3%2F2024

amaanq:
yeah that's fine, i've just invited him

eed3si9n:
Thanks!!

PR #631, opened 2026-08-06
fix: end an operator where a comment begins

exoego:
SLS 1.1 stops an operator at a `/` that opens a comment, so `1 +/*c*/ 2` adds two numbers. The operator tokens took `/` as an ordinary character and swallowed the opener, which built a wrong tree with no error to show for it. Every `/` inside an operator is now bound to the character after it, which rules out `//` and `/*`.

A regex cannot see past the last character it matches, so it can never tell an operator ending in `/` from one that stops before a comment. The nine left-associative classes and the union name token are external as well, and the scanner claims exactly those operators. Each keeps its grammar definition, which the lexer falls back to for everything else, so no state gains a column. The `_`-prefixed tail of an alphanumeric name follows the same rule, with `/` allowed as the whole tail so that names like `Int_/` still read.

Seen in https://github.com/scala/scala3/blob/a036a3a74024d97cc70b8d51f2612c05ec7ff881/tests/semanticdb/expect/Synthetic.expect.scala#L6
Seen in https://github.com/scala/scala3/blob/a036a3a74024d97cc70b8d51f2612c05ec7ff881/tests/pos/i7119.scala#L3


eed3si9n (approved):
Thanks!

PR #636, opened 2026-08-07
fix: keep a `//>` comment plain when no using directive follows

exoego:
The `>` after `//` was an internal token whose precedence beat the comment text, so every `//>` opened a using directive. A line without `using` after it then failed to parse. A regex cannot look past the `>`, so the scanner now claims it only when `using` really follows and otherwise leaves the whole line to the comment text.

Seen in https://github.com/scala/scala3/blob/a036a3a74024d97cc70b8d51f2612c05ec7ff881/tests/run/t7278.scala#L19 Seen in https://github.com/scala/scala3/blob/a036a3a74024d97cc70b8d51f2612c05ec7ff881/tests/run-macros/annot-mod-class-mod-def/Test_2.scala#L5
Seen in https://github.com/scala/scala3/blob/a036a3a74024d97cc70b8d51f2612c05ec7ff881/tests/warn/tostring-interpolated.scala#L2


eed3si9n (approved):
Thanks!

PR #637, opened 2026-08-07
fix: end a statement before a leading operator with no operand

exoego:
A line starting with a symbolic operator continues the line before it, but only when the token after the operator can begin an expression. The gate asked whether anything at all followed, so `??? match` joined the previous statement and then failed on the `match`, and a `???` before a closing brace swallowed the brace.

The word test is the one the capture checking modifiers already use, now shared under its own name. A closing delimiter or a list separator is no operand either, which the postfix reading tested for itself and can now leave to the shared check.

Seen in https://github.com/scala/scala3/blob/a036a3a74024d97cc70b8d51f2612c05ec7ff881/tests/pos/t8300-patmat-a.scala#L16 Seen in https://github.com/scala/scala3/blob/a036a3a74024d97cc70b8d51f2612c05ec7ff881/tests/warn/nonunit-statement.scala#L45


eed3si9n (approved):
Thanks!

PR #641, opened 2026-08-08
fix: annotate a literal type and every annotation of an ascription

exoego:
`ascription_expression` took a single annotation. SLS 6.13 reads
`Ascription ::= ':' Annotation {Annotation}`, so `(c: @switch @unchecked)`
failed to parse.

`annotated_type` wanted a simple type on its left, and a literal type is not
one of those, so `val x: "abc" @deprecated` failed as well.


eed3si9n (approved):
Thanks!

PR #642, opened 2026-08-08
fix: end an indented argument at the comma before the next one

exoego:
A comma only ended an indented block when it ended its line, so an argument written as an indented block failed to parse when the next argument followed on the same line. The comma now also ends the block when the rest of the line closes a bracket that was already open. `import a.b, c.d` and `extends A, B` reach no such closer and keep the old reading.

Seen in https://github.com/slick/slick/blob/a9756aa9e12730a1e8660a6f39bba53c3c58d302/slick/src/main/scala/slick/jdbc/HsqldbProfile.scala#L237
Seen in https://github.com/scala/scala3/blob/a036a3a74024d97cc70b8d51f2612c05ec7ff881/tests/pos/indent-in-parens.scala#L30

### DIFF
--- a/src/scanner.c
+++ b/src/scanner.c
@@ -420,6 +420,50 @@ static bool consume_block_comment_body_on_line(TSLexer *lexer) {
   return consume_block_comment_body_ex(lexer, true);
 }
 
+// Whether the rest of the line closes a bracket that was already open. Such a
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Thanks!

PR #649, opened 2026-08-08
fix: take an nameless erased parameter

exoego:
The reference parser reads `erased` once, right after the paren of a function type or of a using clause, and it applies to a parameter written as a bare type. Only the named spelling was accepted, so `(erased Int) => Int` and `(using erased Int)` failed to parse.

Keeping the modifier out of the comma repeat matters. Inside it the tuple reading of the same parens loses its GLR version and a match type written as a tuple element stops parsing.

Seen in https://github.com/scala/scala3/blob/7d4833b619d31ea9acac97fccf1e74b42b89c49f/tests/pos/i6009a.scala#L4 
Seen in https://github.com/scala/scala3/blob/7d4833b619d31ea9acac97fccf1e74b42b89c49f/tests/pos/CanThrow.scala#L6 
Seen in https://github.com/scala/scala3/blob/7d4833b619d31ea9acac97fccf1e74b42b89c49f/tests/pos/erased-soft-keyword.scala#L14

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -1416,7 +1416,9 @@ module.exports = grammar({
               "using",
               choice(
                 trailingCommaSep1($.class_parameter),
-                trailingCommaSep1($._param_type),
+                // `erased` sits once after `using`. The named parameters carry
+                // their own, so only the unnamed types take it here.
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Thanks!

PR #650, opened 2026-08-08
fix: take the reach capability in a capture set

exoego:
A capture set member may carry a `*`, which stands for what the elements of that capability capture. The star arrives as two different tokens. Right before the closing brace it is the external one that the postfix reading uses, so `->{xs*}` needs that spelling while `->{x, xs*}` and `^{futs*}` get the shared string token.

Seen in https://github.com/scala/scala3/blob/7d4833b619d31ea9acac97fccf1e74b42b89c49f/tests/pos-custom-args/captures/reaches.scala#L31
Seen in https://github.com/scala/scala3/blob/7d4833b619d31ea9acac97fccf1e74b42b89c49f/tests/pos-custom-args/captures/path-use.scala#L16
Seen in https://github.com/scala/scala3/blob/7d4833b619d31ea9acac97fccf1e74b42b89c49f/tests/pos-custom-args/captures/boxed-use.scala#L4

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -1685,6 +1685,10 @@ module.exports = grammar({
     capture_ref: $ =>
       seq(
         choice($._identifier, $.stable_identifier, "cap"),
+        // The reach capability `xs*`, which stands for what the elements of
+        // `xs` capture. It comes before the narrowing suffixes. Right before
+        // the closing brace the star arrives as the external token instead.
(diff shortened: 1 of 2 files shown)


eed3si9n:
Unless the pull request goes unreviewed for a while, could we try to get a review for user-facing changes please?

exoego:
Sorry, understood.

PR #656, opened 2026-08-09
fix: close seven indentation and Scala 2 gaps

exoego:
Each of these reads a construct that the reference parser accepts. The added tests show the shapes.

Three of them need more lookahead than the parser has. Each is answered by giving the scanner a way to see what the parser cannot.

A comma that ends an indented argument asks whether an operand just ended. Only then does it separate arguments and close the block. A list of names offers no operator there, so it is left alone. The closer on the rest of the line settles what remains, where no operand ended before the comma.

The `case` that opens a definition, and the line break Scala disables inside a def header, arrive as tokens of their own. The two readings then reach the parser as different tokens rather than as one it has to guess about.

Measured over dotty 3.8.4 and a corpus of 35 repositories. Failures drop from 60 to 48 and from 430 to 412, no file parses worse, and the parse tables lose 27 states.

Seen in https://github.com/scala/scala3/blob/7d4833b619d31ea9acac97fccf1e74b42b89c49f/tests/run/i6664/Bomb_1.scala#L2 
Seen in https://github.com/scala/scala3/blob/7d4833b619d31ea9acac97fccf1e74b42b89c49f/tests/pos/indent4.scala#L8 Seen in https://github.com/scala/scala3/blob/7d4833b619d31ea9acac97fccf1e74b42b89c49f/tests/pos/trailingCommas/trailingCommas.scala#L64 
Seen in https://github.com/scala/scala3/blob/7d4833b619d31ea9acac97fccf1e74b42b89c49f/tests/pos/indent.scala#L96 
Seen in https://github.com/scala/scala3/blob/7d4833b619d31ea9acac97fccf1e74b42b89c49f/tests/pos/i22527.scala#L51 
Seen in https://github.com/scala/scala3/blob/7d4833b619d31ea9acac97fccf1e74b42b89c49f/tests/pos/i21749.scala#L6 
Seen in https://github.com/scala/scala3/blob/7d4833b619d31ea9acac97fccf1e74b42b89c49f/tests/pos-scala2/i871.scala#L3
Seen in https://github.com/scala/scala3/blob/7d4833b619d31ea9acac97fccf1e74b42b89c49f/tests/pos-scala2/doWhile.scala#L5

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -211,6 +211,11 @@ const XML_NAME = /[_\p{L}][-.:_\p{L}\p{Nd}]*/;
 // such a colon must handle it.
 const colonEol = $ => choice(":", alias($._colon_eol, ":"));
 
+// Inside a case clause body the parser cannot tell this `case` from the one
+// that opens the next clause. The scanner reads past the word and can.
+const caseKeyword = $ =>
+  choice("case", alias($._case_definition_keyword, "case"));
+
 const ascriptionArrowTail = $ =>
   seq(anyArrow(), field("return_type", $._param_type));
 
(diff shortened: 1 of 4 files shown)


eed3si9n (approved):
Thanks!

PR #660, opened 2026-08-10
fix: close two polymorphic type gaps

exoego:
1. A polymorphic function body that is not a parameter clause did not parse. `[T] => ((x: T) => x)` and `[T] => { (x: T) => x }` both ended in an error node.
    - The reference grammar takes any expression after the type parameters, so a branch for the parenthesized and braced bodies is added. The braced one ties with a capture set and takes a conflict.
2. A wildcard bounded by a type lambda did not parse.
    - `? <: [X, Y] =>> Map[X, Y]` reaches the parser as an infix type, whose operand admits no type lambda. Admitting one in every operand position doubles the type sublanguage, so the scanner marks the `?` of a bounded wildcard instead and the lambda is reached only from that token. The tree keeps the infix type shape.

Measured over dotty 3.8.4 and a corpus of 52,236 files. Failures drop from 48 to 46. Five trees change, which are the two fixed files in both checkouts of the reference tests and one more that still fails and now recovers over a shorter region.

Seen in https://github.com/scala/scala3/blob/7d4833b619d31ea9acac97fccf1e74b42b89c49f/tests/run/polymorphic-functions.scala#L106
Seen in https://github.com/scala/scala3/blob/7d4833b619d31ea9acac97fccf1e74b42b89c49f/tests/run/enum-values.scala#L87

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -514,6 +514,8 @@ module.exports = grammar({
     // 'if'  parenthesized_expression  •  '{'  …
     [$._if_condition_paren, $._simple_expression],
     [$.block, $._braced_template_body1],
+    // type_parameters  '=>'  '{'  '}'  •  '['  …
+    [$.block, $.capture_set],
     // '{'  identifier  •  ':' starts the braced typed lambda, the block
     // lambda param, and a statement alike.
     [$._simple_expression, $._braced_typed_lambda],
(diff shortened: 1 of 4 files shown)


eed3si9n (approved):
Thanks!

PR #662, opened 2026-08-10
fix: read a constructor annotation the way Dotty does

exoego:
`class Baz @deprecated(implicit c: C)` and
`class Bar @publicInBinary(@publicInBinary private[Bar] val x: Int)` ended in an error node, and `class Bam @ann(obj)(obj, "h")(n: String)` did too because only one argument list was allowed.

The parenthesis after a constructor annotation opens either an argument list or the class parameters. Whitespace used to decide, which is not what the reference parser does. It reads an empty list as an argument list and hands the parenthesis to the parameters as soon as they could hold what is inside, and the two weights say the same thing.

The name drops the applied constructor type, which was taking the parenthesis into the name and splitting an empty argument list into two nodes. That is why some files in the corpus report a different tree. Failures drop from 46 to 44 in dotty 3.8.4

Seen in https://github.com/scala/scala3/blob/7d4833b619d31ea9acac97fccf1e74b42b89c49f/tests/pos/i2426.scala#L5 
Seen in https://github.com/scala/scala3/blob/7d4833b619d31ea9acac97fccf1e74b42b89c49f/tests/run/publicInBinary/Lib_1.scala#L125

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -514,6 +514,10 @@ module.exports = grammar({
     // 'if'  parenthesized_expression  •  '{'  …
     [$._if_condition_paren, $._simple_expression],
     [$.block, $._braced_template_body1],
+    // 'class'  identifier  '@'  _type_identifier  •  '('  …
+    [$._constructor_annotation],
+    [$._constructor_annotation_arguments, $.arguments],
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Thanks!

PR #664, opened 2026-08-11
fix: take `implicit` in front of a parenthesized lambda parameter list

exoego:
`{ implicit (x: Int) => x + x }` ended in an error node. The keyword was only read in front of a single unparenthesized parameter.

The parenthesized list moves next to that parameter rather than staying its own branch, so the keyword is read in the state that already expects it. The other spelling costs the parse tables a second copy of the lambda head.

Scala 2 wrote a context function this way. The reference parser still reads it and reports a migration warning.

Failures drop from 44 to 42 in dotty 3.8.4, and from 907 to 901 in my local corpus of 52,236 files.

Seen in https://github.com/scala/scala3/blob/7d4833b619d31ea9acac97fccf1e74b42b89c49f/tests/pos/t3672.scala#L2 
Seen in https://github.com/scala/scala3/blob/7d4833b619d31ea9acac97fccf1e74b42b89c49f/tests/warn/i9266.scala#L5

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -2092,8 +2092,11 @@ module.exports = grammar({
     method_value: $ =>
       prec.left(PREC.call, seq($._simple_expression, $.wildcard)),
 
+    // The parenthesized list rides along so that `implicit` is read in the
+    // state that already expects it. Scala 2 spelled a context function this
+    // way, and the reference parser still reads it.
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Thanks!

PR #666, opened 2026-08-13
fix: read a self type and a chained body in a given

exoego:
- Supersedes https://github.com/tree-sitter/tree-sitter-scala/pull/403

----

`given a: Show[Int] with { self => ... }` read `self` as a lambda parameter, and the typed spelling `{ self: Show[T] => ... }` ended in an error node. The braced branch of the with body was a bare block while the indented branch already took a self type. It now shares the rule the other braced template bodies use, so one state reads the self type in both places.

`given g: Show[Int] with Read[Int] with { ... }` read the brace as a refinement on a missing type, and the indented spelling ended in an error node. The alternative that takes more than one constructor had no body at all. It now ends with an optional body, which is the last clause of the production in the reference grammar.

Couldn't find any real example, so I compiled the codes with scalac 3.8.4 to confirm they are accepted.
image

The trees of all 52,236 files in my local corpus are unchanged.

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -1014,7 +1014,7 @@ module.exports = grammar({
           PREC.control,
           seq($._indent, optional($.self_type), $._block, $._outdent),
         ),
-        seq("{", optional($._block), "}"),
+        seq("{", optional($._braced_template_body1), "}"),
       ),
 
     _extension_template_body: $ =>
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Thanks!

PR #668, opened 2026-08-13
fix: read a lone asterisk as a type operator

exoego:
`val t: (2 * 7 + 1) % 10 = 5` and `(a.type * b.type) =:= (b.type * a.type)`, which are _compile-time operations_, ended in an error node. A repeated parameter makes the shared `*` token valid after a type, and that string token beats the operator regex, so inside parentheses the `*` was always taken as the repeat marker.

The infix type now admits the same token as its operator, the way the infix expression already does. The two readings fork and the one that cannot go on dies, so a `*` before a closing paren stays a repeated parameter and a `*` with an operand after it is the operator. The reference parser splits the same way, by looking at the token that follows the `*`.

Seen in https://github.com/scala/scala3/blob/7d4833b619d31ea9acac97fccf1e74b42b89c49f/tests/pos/singleton-ops-composition.scala#L6
Seen in https://github.com/scala/scala3/blob/7d4833b619d31ea9acac97fccf1e74b42b89c49f/tests/pos/7512.scala#L8

### DIFF
--- a/grammar.js
+++ b/grammar.js
@@ -1783,7 +1783,13 @@ module.exports = grammar({
         seq(
           // SimpleType1 admits a bare Refinement: `A & { type X = Int }`.
           field("left", choice($._infix_type_choice, $._structural_type)),
-          field("operator", wordOrOpName($)),
+          // A repeated parameter makes the shared `*` token valid after a type,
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):
Thanks!

PR #669, opened 2026-08-13
fix: read a binder in a splice pattern

exoego:
`case '{ foo(${bound@MyMatcher(x)}) } =>` ended in an error node. A splice inside a quoted pattern holds a pattern, and the grammar read its body as a block, which has no `@` binder.

The binder now sits in the splice body and in the argument list. The block still reads every other shape, so an extractor in a splice stays a call and its trees do not move. Listing the extractor as a pattern instead breaks the far more common `${ Foo.bar }`, because the head then takes the type flavor of its identifier and the expression reading dies with it. Nothing tells the parser it is inside a quoted pattern, so the binder is admitted in every argument list.

Seen in https://github.com/scala/scala3/blob/7d4833b619d31ea9acac97fccf1e74b42b89c49f/tests/pos-macros/splice-pat/Macro_1.scala#L14
Seen in https://github.com/scala/scala3/blob/7d4833b619d31ea9acac97fccf1e74b42b89c49f/tests/pos-macros/splice-pat.scala#L15
Seen in https://github.com/scala/scala3/blob/7d4833b619d31ea9acac97fccf1e74b42b89c49f/tests/neg-macros/i14696.scala#L8


eed3si9n (approved):
Thanks!

PR #671, opened 2026-08-14
fix: do not read a symbolic keyword as a leading infix operator 

exoego:
`case s: String if g:` with an indented body and then `=> s` on its own line ended in an error node, and so did a colon argument whose colon sits alone on its line. The scanner read every run of operator characters at the start of a line as a continuation of the line above, so the indented body never closed in front of `=>` and the lone `:` got a statement break in front of it.

A symbolic keyword lexes as its own token and never as an identifier, so the reference scanner never reads one as a leading infix operator. Such a line now closes the indented regions above it and still takes no statement break, because it cannot start a statement. The one that can is `@`, which opens an annotation.
Refer https://docs.scala-lang.org/scala3/reference/syntax.html#regular-keywords

Seen in https://github.com/scala/scala3/blob/7d4833b619d31ea9acac97fccf1e74b42b89c49f/tests/pos/i25002.scala#L6
Seen in https://github.com/scala/scala3/blob/7d4833b619d31ea9acac97fccf1e74b42b89c49f/tests/rewrites/i22731b.scala#L44


src/scanner.c
```diff
@@ -752,16 +752,50 @@ static bool operand_follows(TSLexer *lexer) {
...
+    }
+    len++;
+    advance(lexer);
+  }
+  op[len < 3 ? len : 3] = '\0';
+  return len;
+}
+
+// A reserved operator is its own token, never an identifier, so scalac's
+// isOperator rejects it and it cannot open a leading infix line.
```
eed3si9n: Thanks for the contribution. I've not heard of a reserved operator. Is there a link to Scala language specification on this?

eed3si9n: Do you mean a symbolic keyword? - https://docs.scala-lang.org/scala3/reference/syntax.html#keywords

> ## Regular keywords
> ```scala
> ....
> 7 :         =         <-        =>        <:        >:        #
> 8 @         =>>       ?=>
> ```

exoego: Yes, a symbolic keyword sounds more natural. By "a reserved operator" I meant that those operators are reserved in Scala's syntax. I also added the `=>>` and `<%` I had missed.

eed3si9n (approved):
Thanks!

